PR #3, opened 2024-09-11
feat: add ecom api

Letushev:
Just a copy of API from the existing template project.
I will get back to env on window later when we will need to make requests from the client side, such as add product to cart, checkout, etc.


tsconfig.json
```diff
@@ -15,7 +15,7 @@
     "forceConsistentCasingInFileNames": true,
     "baseUrl": ".",
     "paths": {
-      "~/*": ["./app/*"]
+      "~/*": ["./src/*"]
```
tetianaman, line 18: I would keep `app`

tetianaman (approved):

PR #8, opened 2024-09-13
Create CategoryLink component

tetianaman:
Allow user to edit `title` and `categorySlug` with Codux Properties Panel: `CategoryLink` component is created.


src/components/category-link/category-link.tsx
```diff
@@ -0,0 +1,18 @@
+import { NavLink } from '@remix-run/react';
+import { ROUTES } from '~/router/config';
+
+type ClassNameFunction = (states: { isActive: boolean; isPending: boolean }) => string;
+
+export interface CategoryLinkProps {
+    categorySlug: string;
```
tetianaman, line 7: maybe just `slug` ?

Letushev (approved):

src/components/category-link/category-link.tsx
Letushev: Both are fine

src/components/category-link/category-link.tsx
```diff
@@ -0,0 +1,16 @@
+import { NavLink, NavLinkProps } from '@remix-run/react';
+import { ROUTES } from '~/router/config';
+
+export interface CategoryLinkProps {
+    categorySlug: string;
+    title: string;
```
alisey, line 6: Accepting title rather than children will make it impossible to use this component to wrap images, for example.

PR #12, opened 2024-09-16
fixes: move title to children in CategoryLink, update URLs in footer links to social networks

tetianaman:
1. Component CategoryLink: prop `title` removed, `children` added.
2. Footer: social network links updated.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -6,81 +6,82 @@
 
 /** @type {import('eslint').Linter.Config} */
 module.exports = {
-  root: true,
-  parserOptions: {
-    ecmaVersion: "latest",
-    sourceType: "module",
-    ecmaFeatures: {
-      jsx: true,
-    },
-  },
-  env: {
-    browser: true,
-    commonjs: true,
-    es6: true,
-  },
-
-  // Base config
-  extends: ["eslint:recommended"],
-
-  overrides: [
-    // React
-    {
-      files: ["**/*.{js,jsx,ts,tsx}"],
-      plugins: ["react", "jsx-a11y"],
-      extends: [
-        "plugin:react/recommended",
-        "plugin:react/jsx-runtime",
-        "plugin:react-hooks/recommended",
-        "plugin:jsx-a11y/recommended",
-      ],
-      settings: {
-        react: {
-          version: "detect",
+    root: true,
+    parserOptions: {
+        ecmaVersion: 'latest',
+        sourceType: 'module',
+        ecmaFeatures: {
+            jsx: true,
         },
-        formComponents: ["Form"],
-        linkComponents: [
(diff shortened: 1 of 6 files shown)


src/components/footer/footer.tsx
```diff
@@ -92,17 +83,32 @@ export const Footer = ({ className }: FooterProps) => {
                 </ul>
                 <ul>
                     <li>
-                        <Link to={ROUTES.home.to()} className={styles.navItem}>
+                        <Link
+                            to="https://www.facebook.com/WixStudio"
+                            className={styles.navItem}
+                            target="blank"
+                            rel="noreferrer"
```
alisey, line 90: Why noreferrer?

src/components/footer/footer.tsx
```diff
@@ -92,17 +83,32 @@ export const Footer = ({ className }: FooterProps) => {
                 </ul>
                 <ul>
                     <li>
-                        <Link to={ROUTES.home.to()} className={styles.navItem}>
+                        <Link
+                            to="https://www.facebook.com/WixStudio"
+                            className={styles.navItem}
+                            target="blank"
```
alisey, line 89: It's not `blank`, it's `_blank`.

src/components/footer/footer.tsx
tetianaman: We don't  know if we can trust the site user wants to link to. I think lets keep referrer private.

tetianaman: There are a lot of resources which advise to add this attribute. All links in Codux with target="_blank" uses it.
I also cannot remove it in the project without [adding eslint rule](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-target-blank.md#rule-details)
Screenshot 2024-09-17 at 09 20 26

alisey: > There are a lot of resources

Those lots of resources at at least 10 years old and apply only to Internet Explorer.

> I also cannot remove it in the project without [adding eslint rule]

Not adding, removing. It's already added and it's completely useless.

> We don't know if we can trust the site user wants to link to. I think lets keep referrer private.

So this has nothing to do with the trust.

alisey (approved):

.eslintrc.cjs
```diff
@@ -6,81 +6,82 @@
...
+                },
+            },
+            extends: [
+                'plugin:@typescript-eslint/recommended',
+                'plugin:import/recommended',
+                'plugin:import/typescript',
+            ],
+            rules: {
+                '@typescript-eslint/no-explicit-any': 'off',
+                'react/jsx-no-target-blank': 'off',
```
tetianaman, line 75: this added, other just formatting

PR #13, opened 2024-09-17
feat: breadcrumbs

Letushev:
Screenshot 2024-09-17 at 10 41 21
Screenshot 2024-09-17 at 10 41 27

Guide: https://remix.run/docs/en/main/guides/breadcrumbs.

- We don't show breadcrums on every page, only products category and product details.
- On product details we have a special case. Product details route is not nested inside category route. But we want to  show the category link in breadcrumbs if a user visits the product page directly from the category page. I do this using history state. So the category link appears only on the client side.

### DIFF
--- a/app/routes/product-details.$productSlug/route.tsx
+++ b/app/routes/product-details.$productSlug/route.tsx
@@ -1,7 +1,7 @@
 import { getEcomApi } from '~/api/ecom-api';
 import styles from './product-details.module.scss';
 import { json, LoaderFunctionArgs } from '@remix-run/node';
-import { Link, useLoaderData } from '@remix-run/react';
+import { useLoaderData } from '@remix-run/react';
 import { ProductPrice } from '~/components/product-price/product-price';
 import { QuantityInput } from '~/components/quantity-input/quantity-input';
 import { useState } from 'react';
@@ -11,9 +11,9 @@ import { Button } from '~/components/button/button';
 import { removeQueryStringFromUrl } from '~/utils';
 import { ShareProductLinks } from '~/components/share-product-links/share-product-links';
 import { Breadcrumbs } from '~/components/breadcrumbs/breadcrumbs';
-import { ROUTES } from '~/router/config';
 import { RouteHandle } from '~/router/types';
 import { CategoryLink } from '~/components/category-link/category-link';
+import { ProductLink } from '~/components/product-link/product-link';
 
 export const loader = async ({ params, request }: LoaderFunctionArgs) => {
     const productSlug = params.productSlug;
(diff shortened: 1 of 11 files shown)


app/routes/product-details.$productSlug/route.tsx
```diff
@@ -27,54 +31,91 @@ export const loader = async ({ params, request }: LoaderFunctionArgs) => {
...
+}
+
+export const handle: RouteHandle<typeof loader, ProductDetailsLocationState> = {
+    breadcrumb: (match, location) => {
+        const fromCategory = location.state?.fromCategory;
+
+        const productLink = (
+            <Link to={ROUTES.productDetails.to(match.data.product.slug!)} state={{ fromCategory }}>
+                {match.data.product.name!}
+            </Link>
```
tetianaman, line 48: What do you think about `ProductLink` component?
We have also some product links on the home page. So user can use this component to sell specific product.

src/components/breadcrumbs/breadcrumbs.tsx
```diff
@@ -0,0 +1,20 @@
+import { useLocation, useMatches } from '@remix-run/react';
+import { RouteMatch } from '~/router/types';
+import styles from './breadcrumbs.module.scss';
+import { ArrowRightIcon } from '../icons';
+
+export const Breadcrumbs = () => {
+    const matches = useMatches() as RouteMatch[];
+    const location = useLocation();
```
alisey, line 8: Will we be able to create a board for this component?

src/components/breadcrumbs/breadcrumbs.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+    return (
+        <div className={styles.breadcrumbs}>
+            {matches
+                .flatMap((match) => match.handle?.breadcrumb?.(match, location) ?? [])
+                .map((breadcrumb, index, arr) => [
+                    breadcrumb,
+                    index !== arr.length - 1 && <ArrowRightIcon className={styles.separatorIcon} />,
```
alisey, line 16: No warning about missing `key` on children?

alisey (approved):

src/components/breadcrumbs/breadcrumbs.tsx
Letushev: Hope so. If not, I will create a dumb version.

src/components/breadcrumbs/breadcrumbs.tsx
Letushev: No, which is weird. I've added React.Fragment with a key

PR #9, opened 2024-09-13
feat: add Hero image

tetianaman:
What is done:
1. HeroImage. It takes a free space in the viewport. I implement it using variables, so the user could change it in Codux.
2. Header styles fixed a little bit.

https://github.com/user-attachments/assets/0b4da061-095f-4033-bd21-1d0c8d3cab8f

PR #14, opened 2024-09-17
UI: use variables for colors and fonts

tetianaman:
Delete the old _theme.scss_ file and use _colors.scss_ and _typography.scss_


src/components/button/button.module.scss
```diff
@@ -3,17 +3,17 @@
     align-items: center;
     justify-content: center;
     padding: 12px 16px;
-    border: 1px solid #17110d;
-    background-color: #17110d;
+    border: 1px solid var(--colorA5);
+    background-color: var(--colorA5);
     font-size: 16px;
     line-height: 1.5;
```
Letushev, line 9: `font: var(--paragraph1)`

app/routes/product-details.$productSlug/product-details.module.scss
```diff
@@ -6,19 +6,21 @@
...
     line-height: 1;
-    color: #767165;
+    color: var(--colorA4);
 }
 
 .price {
     margin-top: 16px;
-    font: 21px / 1.3 'Marcellus', serif;
+    font: var(--heading5);
+    font-size: 21px;
```
Letushev, line 23: I think we should not override font variables. What's the point of having theme then.
If there is no such variable choose the closest one or adjust/add variables according to the values we use. Exceptions may exist, but it seems to me there are too many of them.

tetianaman: I agree and I don't like it as well. 
I'm against of choosing the closest one. IMO 1px in design matters.
Adjust/add variables - I would do this.

src/components/cart-icon/cart-icon.module.scss
```diff
@@ -9,13 +9,13 @@
 .handle {
     width: 9px;
     height: 7px;
-    border: 1px solid #000;
+    border: 1px solid var(--black);
```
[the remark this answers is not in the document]
tetianaman: I've added

app/routes/product-details.$productSlug/product-details.module.scss
Letushev: 1px matters when we have a design. But we deal with values from the complete website. Sometimes I saw font-size like this:
```css
--scaling-factor: min((100* 1cqw)), 9999px);
font-size: max(0.5px, 0.0195313 * (var(--scaling-factor) - var(--scrollbar-width))) !important;
```
I don't think we want this in out template so I took an approximate value.

tetianaman: Right, when we know exact value. 
In the case you mentioned I use some dynamic values like `5vw` and see how it works on window resize. 
If there are a lot of places we use it, it would be good to have it in variable...

@alisey Do you agree to fit some variable values to the values we use in the project? 
Also If I see a font-size 1-2px different from variable font-size, I can try to fit it.

src/components/cart-icon/cart-icon.module.scss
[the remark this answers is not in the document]
tetianaman: Yeap, its in [different PR ](https://github.com/codux-templates/e-commerce-reclaim/pull/9/files#diff-8454f4f2139e0aeebd5f7e56843fabad729cbbf65a5c9b54c26ef5630fe44bcf)
I'll add it here as well.

app/routes/product-details.$productSlug/product-details.module.scss
alisey: > Do you agree to fit some variable values to the values we use in the project?

I didn't understand the question.

tetianaman: @Letushev We discussed the problem with @alisey and decided to optimise  values and variables after the entire UI is ready.

src/components/cart-icon/cart-icon.module.scss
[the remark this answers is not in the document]
tetianaman: Merged. `--white` and `--black` is in project.

Letushev (approved):

PR #18, opened 2024-09-18
feat: add category links to the home page

tetianaman:
https://github.com/user-attachments/assets/89856c49-e664-4ad3-bf0d-f34e9dd02974


alisey (approved):

Letushev:
Why not framer-motion and components that @alisey prepared?

PR #20, opened 2024-09-19
feat: run full app in codux

ynesteruk:
Adds define app config that allows to run a template in full app mode in Codux.

https://github.com/user-attachments/assets/b9ae3843-277a-460b-9f45-b46479a22e67

## **Important note:**
`define-remix-app` wrapper runs all loaders in web workers, it means we can't use node specific login there (e.g. we can't use things from '@remix-run/node' package).

____

## **Known issue:**
`define-remix-app` doesn't pass correctly an error from loader to a route component, that can lead to weird errors

https://github.com/user-attachments/assets/5e61a4d4-5812-4e2a-b527-1c3fa1c5f557

This error appears because `useLoaderData` returns an empty object instead of an error that a category doesn't exist 

## How to run:
At the moment, we need to use `peter/production-app` branch in Codux and run it with a command:
```bash
npm run codux:electron:release
```

### DIFF
--- a/app/root.tsx
+++ b/app/root.tsx
@@ -7,7 +7,6 @@ import {
     ScrollRestoration,
     useLoaderData,
 } from '@remix-run/react';
-import { json } from '@remix-run/node';
 import { SiteWrapper } from '~/components/site-wrapper/site-wrapper';
 import { ROUTES } from '~/router/config';
 import { RouteHandle } from '~/router/types';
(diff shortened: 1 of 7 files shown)


app/root.tsx
```diff
@@ -16,11 +15,11 @@ import { CartOpenContextProvider } from '~/components/cart/cart-open-context';
 import '~/styles/index.scss';
 
 export async function loader() {
-    return json({
+    return {
```
tetianaman, line 18: why `json` was here?

ynesteruk: With `json` from '@remix-run/node' app won't run in codux because of a problem I wrote in a description.

tetianaman (approved):

PR #23, opened 2024-09-19
feat: thank you page

Letushev:
Based on e-commerece-remix "Thank you" page, but I cleaned it up a bit and adjusted UI to our template.

localhost_5173_thank-you_wixMemberLoggedIn=false orderId=fdf98862-7e16-4bae-858b-9405d3dd73d2 (1)

### DIFF
--- a/app/routes/thank-you/route.tsx
+++ b/app/routes/thank-you/route.tsx
@@ -1,3 +1,31 @@
+import { useLoaderData, json } from '@remix-run/react';
+import type { LoaderFunctionArgs } from '@remix-run/node';
+import { getEcomApi } from '~/api/ecom-api';
+import { CategoryLink } from '~/components/category-link/category-link';
+import { OrderSummary } from '~/components/order-summary/order-summary';
+import styles from './thank-you.module.scss';
+
+export const loader = async ({ request }: LoaderFunctionArgs) => {
+    const orderId = new URL(request.url).searchParams.get('orderId');
+    if (!orderId) throw new Error('Order not found');
+    const order = await getEcomApi().getOrder(orderId);
+    return json({ order });
+};
+
 export default function ThankYouPage() {
-    return <h1>Thank you</h1>;
+    const { order } = useLoaderData<typeof loader>();
+
+    return (
+        <div className={styles.root}>
+            <h1 className="heading4">Thank You!</h1>
+            <div className={styles.subtitle}>{`You'll receive a confirmation email soon.`}</div>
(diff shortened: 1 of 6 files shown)


app/routes/thank-you/route.tsx
```diff
@@ -1,3 +1,31 @@
...
+import { getEcomApi } from '~/api/ecom-api';
+import { CategoryLink } from '~/components/category-link/category-link';
+import { OrderSummary } from '~/components/order-summary/order-summary';
+import styles from './thank-you.module.scss';
+
+export const loader = async ({ request }: LoaderFunctionArgs) => {
+    const orderId = new URL(request.url).searchParams.get('orderId');
+    if (!orderId) throw new Error('Order not found');
+    const order = await getEcomApi().getOrder(orderId);
+    return json({ order });
```
tetianaman, line 12: Do we need `json`?

Letushev: Yes. I import it from `@remix-run/react` now, not `@remix-run/node`, so there shouldn't be a problem.
> This is a shortcut for creating `application/json` responses. Converts `data` to JSON and sets the `Content-Type` header.

ynesteruk: @Letushev anyway it's better to check it in Codux

app/routes/thank-you/route.tsx
```diff
@@ -1,3 +1,31 @@
...
+};
+
 export default function ThankYouPage() {
-    return <h1>Thank you</h1>;
+    const { order } = useLoaderData<typeof loader>();
+
+    return (
+        <div className={styles.root}>
+            <h1 className="heading4">Thank You!</h1>
+            <div className={styles.subtitle}>{`You'll receive a confirmation email soon.`}</div>
```
tetianaman, line 21: why in backticks?

tetianaman (approved):

app/routes/thank-you/route.tsx
Letushev: Because of `'`

tetianaman: Doesn't double quotes work? 
"You'll receive a confirmation email soon."

Is the difference in the Codux: text node in "" and text node based on {} ?

Letushev: Double quotes no.
There is no big diff in Codux when text node is wrapped in {``} or not, only in the props panel it shows `Edit code` instead of text input.
I'll escape it with \&apos;

ynesteruk (approved):

PR #29, opened 2024-09-20
feat: about page

ynesteruk:
Adds about page

Screenshot 2024-09-20 at 15 39 42

____

Screenshot 2024-09-20 at 15 40 41

____

Screenshot 2024-09-20 at 15 41 09

### DIFF
--- a/app/routes/about-us/about-us.module.scss
+++ b/app/routes/about-us/about-us.module.scss
@@ -1,0 +1,74 @@
+.root {
+    display: grid;
+    grid-template-columns: 1fr 1fr;
+    min-height: 85vh;
+    margin-bottom: 60px;
+    gap: 6px;
+}
+
+.aboutSection {
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+}
+
+.header1 {
+    font: var(--paragraph1);
+    font-size: 22px;
+    margin-bottom: 8px;
+}
+
+.header2 {
+    font: var(--heading2);
+    font-size: 60px;
+    max-width: 600px;
+    margin-bottom: 8%;
+}
+
+.description {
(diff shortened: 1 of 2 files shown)


app/routes/about-us/about-us.module.scss
```diff
@@ -0,0 +1,74 @@
...
+
+.aboutSection {
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+}
+
+.header1 {
+    font: var(--paragraph1);
+    font-size: 22px;
```
Letushev, line 17: I think `font` declaration can be removed. You defined custom font-size, line-height and font-family are inherited from root.

Letushev (approved):

app/routes/about-us/route.tsx
```diff
@@ -1,3 +1,22 @@
+import styles from './about-us.module.scss';
+
 export default function AboutUsPage() {
-    return <h1>About us</h1>;
+    return (
+        <div className={styles.root}>
+            <div className={styles.aboutSection}>
+                <h1 className={styles.header1}>We are ReClaim</h1>
+                <p className={styles.header2}>A women-owned local business</p>
```
alisey, line 8: Call these `title` and `subtitle`. Because we have common styles called `heading1` and `heading2`, and it's too confusing when the styles don't match those.

app/routes/about-us/route.tsx
```diff
@@ -1,3 +1,22 @@
...
+                <p className={styles.description}>
```
alisey, line 9: Don't use `<p>` elements. They don't allow putting block elements inside of them, and it's generally not something that people using a visual editor can understand or handle correctly.

alisey (approved):

PR #32, opened 2024-09-20
feat: responsive product page

Letushev:
https://github.com/user-attachments/assets/0947a861-0542-4cea-a410-32671a0eb5e9

- make `vite.config.ts` the same as in [e-commerce-remix](https://github.com/codux-templates/e-commerce-remix/blob/9dddba82ed017bb7e8020390fca1ab5d1cc00625/vite.config.ts)
- p.s. oveflow hidden is applied for header and footer in video

### DIFF
--- a/src/styles/general.scss
+++ b/src/styles/general.scss
@@ -1,0 +1,3 @@
+$mobile-width: 480px;
+$tablet-width: 768px;
+$desktop-width: 1024px;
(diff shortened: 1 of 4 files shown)


src/styles/general.scss
```diff
@@ -0,0 +1,3 @@
+$mobile-width: 480px;
+$tablet-width: 768px;
+$desktop-width: 1024px;
```
alisey, line 3: Is there any meaningful difference between `common.scss` and `general.scss`? As a user I would not understand it. Maybe let's put these in `common.scss`?

alisey (approved):

PR #58, opened 2024-09-23
feat: make products category page responsive + add product card animation

Letushev:
https://github.com/user-attachments/assets/e0fac5b8-7a4f-4414-8695-0872b526cc9c

- Hide categories navigation on small viewports
- I can't find the exact animation duration so I set the duration that feels the most accurate to me.

### DIFF
--- a/app/routes/products.$categorySlug/products.module.scss
+++ b/app/routes/products.$categorySlug/products.module.scss
@@ -1,3 +1,5 @@
+@import '@styles/common.scss';
+
 .page {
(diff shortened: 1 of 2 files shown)


app/routes/products.$categorySlug/products.module.scss
```diff
@@ -1,3 +1,5 @@
+@import '@styles/common.scss';
```
alisey, line 1: Remove

alisey (approved):

PR #71, opened 2024-09-23
feat: responsive header

Letushev:
https://github.com/user-attachments/assets/6c43b36f-6f9d-4322-aafd-af7a1eee4257

- Improve drawer: add animation using framer motion, render in portal.
- Delete hidden login icon.


alisey (approved):

src/components/sidebar-navigation-menu/sidebar-navigation-menu.module.scss
```diff
@@ -0,0 +1,29 @@
...
+    cursor: pointer;
+}
+
+@media (max-width: $tablet-width) {
+    .menu {
+        font: var(--paragraph1);
+    }
+
+    .drawer {
+        max-width: none;
```
alisey, line 27: What does this do?

alisey: I see.

src/components/navigation-menu/navigation-menu.tsx
```diff
@@ -0,0 +1,74 @@
...
+                        onClick={onLinkClick}
+                    >
+                        Shop All
+                    </CategoryLink>
+                </li>
+                <li>
+                    <CategoryLink
+                        categorySlug="kitchen-essentials"
+                        className={menuItemStyle}
+                        onClick={onLinkClick}
```
alisey, line 39: Does remix give us a hook something like 'onNavigate'? Or maybe we can add a single click handler on the root and check if event target is a link. So if the user wants to add/remove an item they wouldn't need to wire it to onClick.

src/components/header/header.module.scss
```diff
@@ -9,27 +10,31 @@
 }
 
 .topBar {
+    min-height: 50px;
```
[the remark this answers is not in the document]
Letushev: Yes, I missed this detail. The top bar has bigger height on mobile

alisey (approved):

src/components/header/header.module.scss
```diff
@@ -1,35 +1,41 @@
+@import '@styles/variables.scss';
+
 :root {
     --headerTopBarHeight: 50px;
     --headerNavigationHeight: 54px;
-    --headerHeight: 104px;
+    --headerHeight: calc(var(--headerTopBarHeight) + var(--headerNavigationHeight));
```
alisey, line 6: The reason we didn't define `--headerHeight` as a calc previously is that the Style Panel shows such values as invalid values (with a strike-through).

PR #75, opened 2024-09-23
fix: getErrorMessage returns an object in case of ErrorResponse

ynesteruk:
#### Problem
In case of `ErrorResponse` emitted by route loader, `getErrorMessage` function return an object instead of a string.

### DIFF
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -39,7 +39,9 @@ export function getErrorMessage(value: unknown): string {
     }
 
     if (isRouteErrorResponse(value)) {
-        return value.data.message;
+        if (typeof value.data === 'string') return value.data;
+        else if (typeof value.data.message === 'string') return value.data.message;
+        else return String(value.data);
     }
 
     if (isEcomSDKError(value)) {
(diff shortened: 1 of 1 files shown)


src/utils/index.ts
```diff
@@ -39,7 +39,7 @@ export function getErrorMessage(value: unknown): string {
     }
 
     if (isRouteErrorResponse(value)) {
-        return value.data;
+        return value.data.message;
```
alisey, line 42: Prolly if (typeof foo === 'string') ... else if (typeof foobar === 'string') ... else String(value.data)

Because the type of data is `any`.

Letushev (approved):

src/utils/index.ts
```diff
@@ -39,7 +39,9 @@ export function getErrorMessage(value: unknown): string {
     }
 
     if (isRouteErrorResponse(value)) {
-        return value.data.message;
+        if (typeof value.data === 'string') return value.data;
+        else if (typeof value.data.message === 'string') return value.data.message;
```
alisey, line 43: Let's immediately cast data to unknown to make sure we're not doing something stupid:

```ts
const data: unknown = value.data;
```

And add a comment:
If the thrown error is an instance of `Error`, React Router will convert it to a string. If it's a JSON response like `throw json({ message: "error" })` it will parse that JSON into `ErrorResponse.data`.

PR #76, opened 2024-09-23
UI: responsive hero banner

tetianaman:
I keep banner height full-screen  for all screens:
`--bannerHeight: calc(100vh - var(--headerHeight) - var(--marginBottom));`


src/styles/common.scss
```diff
@@ -142,6 +142,66 @@
...
+        margin-left: 2vw;
+        margin-right: 2vw;
+    }
+}
+
+@media (max-width: $tablet-width) {
+    .heroBanner {
+        --marginBottom: 8vw;
+        margin-left: 6vw;
+        margin-right: 6vw;
```
Letushev, line 201: I don't understand why the hero banner gets thinner as the viewport becomes smaller. Maybe let's not add side margins at all?

Letushev (commented):
I see in the original UI it takes around 50vh from $tablet-width to $desktop-width

Screenshot 2024-09-23 at 18 36 30

Letushev (approved):

src/styles/common.scss
tetianaman: I've put margins to all home page section by wix template design so far.

tetianaman:
> I see in the original UI it takes around 50vh from $tablet-width to $desktop-width

I decided to fit it to screen, because I don't know what height to take. I think it should depends on some tablet dimensions to look good. To keep it stretched much easier and looks good.

PR #78, opened 2024-09-23
feat: responsive text banner

tetianaman:
https://github.com/user-attachments/assets/5aba6cae-2a8d-47f0-96c6-dfb3c46fb0c6


tetianaman:
I run prettier, it doesn't make any changes in  _common.scss_ file.

alisey:
@tetianaman "No description provided."

src/styles/common.scss
```diff
@@ -236,6 +236,36 @@
...
+        padding: 70px 0;
+    }
+
+    .textBannerSubtitle {
+        font-size: 14px;
+    }
+
+    .textBannerTitle {
+        font-size: 6.3vw;
+    }
```
alisey, line 266: The min size needs to be limited because it goes to the same size as subtitle as you shrink the window.

src/styles/common.scss
```diff
@@ -236,6 +236,36 @@
...
 }
 
+@media (max-width: $desktop-width) {
+    .textBanner {
+        padding: 70px 14%;
+    }
+
+    .textBannerSection {
+        margin-left: 2vw;
+        margin-right: 2vw;
```
alisey, line 246: What is this for?

src/styles/common.scss
```diff
@@ -236,6 +236,37 @@
...
+    .textBannerSubtitle {
+        font-size: 14px;
+    }
+
+    .textBannerTitle {
+        font-size: 6.3vw;
+    }
+}
+
+@media (max-width: $tablet-width) {
```
alisey, line 264: You already have a tablet width media query above.

src/styles/common.scss
tetianaman: To avoid this:
Screenshot 2024-09-24 at 14 22 39
Let me go around this

alisey: And what exactly are we looking to avoid here?

tetianaman: Actually I need that, but for mobile
Screenshot 2024-09-24 at 14 30 59
You can see on wix template padding 4%. 
We have 2vw on <main>. I need additional 2vw to get that 4%

src/styles/common.scss
tetianaman: I see it. Fixed

src/styles/common.scss
tetianaman: > And what exactly are we looking to avoid here?

That text is too close to margins. We need to have some space there

alisey (approved):

PR #79, opened 2024-09-23
feat: responsive categories section on the home page

tetianaman:
Desktop and tablet:
Screenshot 2024-09-24 at 20 25 30

Mobile:
Screenshot 2024-09-24 at 20 29 30


src/styles/common.scss
```diff
@@ -265,5 +266,12 @@
     gap: 4px;
     width: 100%;
     margin-bottom: 5%;
-    aspect-ratio: 2.26;
+}
+
+@media (max-width: $tablet-width) {
+    .rowOfCards {
+        grid-auto-rows: minmax(0, 1fr);
+        grid-auto-flow: row;
```
tetianaman, line 274: I stay with `grid`, because `display: block` doesn't give me `gap` between cards.

alisey: I don't understand.

src/styles/common.scss
```diff
@@ -178,6 +178,7 @@
 
 .linkCard {
     position: relative;
+    aspect-ratio: 0.75;
```
alisey, line 181: Why?

tetianaman: to remove `aspect-ratio: 2.26;` for entire row and to not set` aspect-ratio: 0.75;` in media queries where cards aligned vertically.

alisey: IMO it doesn't make sense to hardcode aspect ratio into a link card.

src/styles/common.scss
alisey: I see what happened here. The actual Wix website built on this template doesn't have any wrapping at mobile width:

image

alisey: It only wraps when you reload the page. So I didn't understand what kind of stying you're trying to achieve.

alisey: I think you should use flex in both cases and change flex-direction from row to column.

alisey: Put aspect-ratio on the direct children of rowOfCards. Rename rowOfCards to something cardGroup.

src/styles/common.scss
```diff
@@ -280,6 +280,20 @@
...
+    }
+}
+
+@media (max-width: $tablet-width) {
+    .linkCardTitle {
+        font-family: var(--text-font);
+        font-size: 22px;
+        font-weight: 600;
+    }
+}
```
alisey, line 295: On mobile the cards are larger than on tablet. I don't see the point of switching to a smaller font for larger cards.

alisey:
I've pushed changes:

1. Removed redundant `width: 100%` from `cardGroup`
2. Moved `cardGroup` to its own section in the CSS file instead of "Miscellaneous"
3. Used a container query to define the font size of `linkCardTitle`. Using viewport units doesn't make sense for this component, because a single row can contain anywhere between 1 and 5-6 cards. So the font size should depend on the width of the card rather than the width of the viewport.

alisey (approved):

PR #80, opened 2024-09-23
UI: responsive featured products on the home page

tetianaman:
https://github.com/user-attachments/assets/7689fa06-a79b-45ee-ba6c-166cbfa70d3d

### DIFF
--- a/src/components/featured-products-section/featured-products-section.module.scss
+++ b/src/components/featured-products-section/featured-products-section.module.scss
@@ -52,9 +52,9 @@
     }
 
     .header {
-       display: block;
-       font-size: 30px;
-       margin-bottom: 40px;
+        display: block;
+        font-size: 30px;
+        margin-bottom: 40px;
     }
 
     .headerTitle {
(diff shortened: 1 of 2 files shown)


tetianaman:
No idea why pretest failes.

src/components/featured-products-section/featured-products-section.module.scss
```diff
@@ -31,3 +33,42 @@
...
+        font-size: 30px;
+        margin-bottom: 16px;
+    }
+
+    .headerDescription {
+        font-size: 14px;
+    }
+
+    .productsRow {
+        grid-template-columns: repeat(auto-fit, minmax(min(135px, 100%), 1fr));
```
alisey, line 70: `repeat(auto-fit, minmax(min(135px, 100%), 1fr))`

Can you explain this?

tetianaman: Right. The expression is good for wix template where card `width` (here 135px) is dynamic and can be bigger than `grid` width. In this case `min(--card-width, 100%)` provides min column width if `--card-width < grid width`
We have no grid less than 135px of witdh, so the expression can be simplified to `repeat(auto-fit, minmax(135px, 1fr));`

alisey (approved):

PR #83, opened 2024-09-24
fix: improve handling of React Router's ErrorResponse in  getErrorMessage

ynesteruk:
* Correctly extract the error message when one of the following patterns is used:
    * `throw new Response('oops')`
    * `throw json('oops')`
    * `throw json({message: 'oops'})`

* Update function description
* Update `ErrorBoundary` for the product page and category page to use `getErrorMessage`.

### DIFF
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -38,10 +38,17 @@ export function getErrorMessage(value: unknown): string {
         return value.message;
     }
 
+    // If the thrown error is an instance of Error, React Router will convert it to a string.
+    // If it's a JSON response like throw json({ message: "error" }) it will parse that JSON into ErrorResponse.data.
     if (isRouteErrorResponse(value)) {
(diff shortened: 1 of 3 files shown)


src/utils/index.ts
```diff
@@ -38,10 +38,17 @@ export function getErrorMessage(value: unknown): string {
         return value.message;
     }
 
+    // If the thrown error is an instance of Error, React Router will convert it to a string.
```
alisey, line 41: Just talked to Oleh, this first line of the comment is not quite correct. This only applies to errors wrapped with React Router's private [ErrorResponseImpl](https://github.com/remix-run/react-router/blob/9c32cfa3529071373ff306d7711c34f7052dfc78/packages/router/utils.ts#L1670) object.

I'll push an updated comment to this branch.

alisey (approved):

PR #90, opened 2024-09-24
feat: show out of stock indication

Letushev:
Show product out of stock indication. I don't handle out of stock variants.

Screenshot 2024-09-24 at 16 07 56
Screenshot 2024-09-24 at 16 08 30
Screenshot 2024-09-24 at 16 18 45

### DIFF
--- a/src/styles/common.scss
+++ b/src/styles/common.scss
@@ -58,6 +58,7 @@
 }
 
 .button:disabled {
+    border-color: transparent;
     background-color: var(--colorA3);
     color: var(--colorA1);
     cursor: default;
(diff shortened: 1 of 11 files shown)


src/styles/common.scss
```diff
@@ -58,6 +58,7 @@
 }
 
 .button:disabled {
+    border-color: transparent;
```
Letushev, line 61: I am not sure how the disabled button should look like. But out of stock button looks like this for sure:

Screenshot 2024-09-24 at 16 22 43

So I removed border when a button is disabled.

Letushev:
When the product is already in cart and then it becomes out of stock, the cart data is not refreshed until a user updates something in cart (changes quality, tries to checkout). If a user tries to checkout, a checkout page will look like this:

Screenshot 2024-09-24 at 16 16 04

 Only after that the cart data is refreshed and we show that the cart item is no longer available.
 This is not related to SWR cache. This is how it works on wix website too.

alisey (approved):

PR #91, opened 2024-09-24
Remove focus shadow from .button:focus

Dana-Levy:
image


tetianaman:
No shadow anymore, check main.

PR #95, opened 2024-09-24
feat: skeletons for the featured products section

alisey:
Since the `<FeaturedProductsSection>` is loaded on the client, sometimes there's an unpleasant page jump if you scroll before this section has loaded.

I've added skeletons that normally won't be visible, you have to scroll down to the products section fast enough. I wasn't sure whether to pass dummy data into the `<ProductCard>` component, or to add a prop `skeleton={true}` or to implement a separate `<ProductCardSkeleton>`.

https://github.com/user-attachments/assets/c2d73c98-95c3-4d64-a70b-9321073f98bb


alisey:
@tetianaman applying a different animation to the skeletons would mean that this animation stops abruptly once the data is a loaded, and the `<Reveal>` animation begins. It would look ugly.

PR #105, opened 2024-09-26
Fix component UI kit and added variants to them

mikawix:
1. Fixed price issue when price is not defined
image
(showed price with a strike through before change)

2. Added missing variants to components
3. Added label variant
4. Added animation kit - variants to animations board, but Parallax, need your help on it

PR #112, opened 2024-09-26
feat: tidy up the components ui kit and add variant wrappers

alisey:
"Components" UI Kit board:
* Removed unused CSS.
* Wrapped snippets with `<Variant>`.
* Fixed incorrect names for existing variants.
* Removed all board-specific CSS/HTML from variants.
* Added discounted price for the product card.
* Added an example of a "SALE" ribbon.
* Introduced a common class for the ribbon element.

screenshot

### DIFF
--- a/_codux/ui-kit/components/components.board.tsx
+++ b/_codux/ui-kit/components/components.board.tsx
@@ -1,5 +1,4 @@
 import { createBoard, Variant } from '@wixc3/react-board';
-import classNames from 'classnames';
 import { Accordion } from '~/components/accordion/accordion';
 import { ProductCard } from '~/components/product-card/product-card';
 import { QuantityInput } from '~/components/quantity-input/quantity-input';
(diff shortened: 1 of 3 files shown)


_codux/ui-kit/components/components.board.tsx
```diff
@@ -17,46 +16,50 @@ export default createBoard({
...
+                <h4 className={styles.sectionHeader}>INPUT</h4>
+                <Variant name="Number Input">
+                    <QuantityInput value={6} onChange={() => {}} />
                 </Variant>
-                <p className={classNames(styles.variantName, styles.headlinesSpacing)}></p>
+                <span className={styles.variantName}>Number Input</span>
+                <hr className={styles.hrLight} />
+                <h4 className={styles.sectionHeader}>ACCORDION</h4>
+                <div style={{ width: '70%' }}>
+                    <Variant name="Accordion">
```
tetianaman, line 28: Why do we wrap components into `Variant` if we have them in `AddPanel` ?

alisey: No idea 🤷‍♂️ Doesn't make sense to me as it creates duplicate instances of the same thing in the panel. But this is what they want in UI Kits.

alisey: This PR is just a cleaned-up version of https://github.com/codux-templates/e-commerce-reclaim/pull/105

PR #114, opened 2024-09-30
feat: add RangeSlider component

Letushev:
https://github.com/user-attachments/assets/3dbde6e3-946b-45a3-b7f6-5720a22e5de4

### DIFF
--- a/src/components/range-slider/range-slider.tsx
+++ b/src/components/range-slider/range-slider.tsx
@@ -28,8 +28,7 @@ interface RangeSlider {
 }
 
 /**
- * A slider that allows selecting a numeric range.
- * It is built from 3 native <input type="range"> elements layered on top of each other.
+ * A slider component for selecting a numeric range.
  */
 export const RangeSlider = ({
     startValue,
@@ -74,6 +73,9 @@ export const RangeSlider = ({
 
     return (
         <div>
+            {/* The slider is implemented using three native <input type="range" />
+                elements stacked on top of each other.
+            */}
             <div className={styles.slidersContainer}>
                 {/* Displays the track with the highlighted selected range. The thumb is hidden.
                     Handles a change when a user clicks somewhere on the track.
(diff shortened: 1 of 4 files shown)


tetianaman (approved):

src/components/range-slider/range-slider.tsx
```diff
@@ -0,0 +1,130 @@
...
+    const handleStartValueChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+        const newStartValue = Number(event.target.value);
+        if (newStartValue <= endValue) {
+            onStartValueChange(newStartValue);
+        }
+    };
+
+    const handleEndValueChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+        const newEndValue = Number(event.target.value);
+        if (newEndValue >= startValue) {
```
alisey, line 54: This should be `max` rather than `if`. Same for the `handleStartValueChange`.

https://github.com/user-attachments/assets/bfef75a0-d4cd-4349-af62-64dd3b1cc035

src/components/range-slider/range-slider.tsx
```diff
@@ -0,0 +1,130 @@
...
+        }
+    };
+
+    const handleChangeByClickingOnTrack = (event: React.ChangeEvent<HTMLInputElement>) => {
+        const value = Number(event.target.value);
+        // Change the start or end value
+        // depending on which one is closer to the clicked value on the track.
+        if (Math.abs(value - startValue) < Math.abs(value - endValue)) {
+            onStartValueChange(value);
+        } else {
```
alisey, line 65: If the min and max are equal, we should choose the one that can be adjusted.

Ideally, we should also track which one was selected and avoid switching to the other during a single drag.

Letushev: > Ideally, we should also track which one was selected and avoid switching to the other during a single drag.

I don't know how to reproduce switching to the other during a single drag.

alisey: > I don't know how to reproduce switching to the other during a single drag.

Here I modified the style of one of the thumbs to show it more clearly. It's not worth fixing, since it requires introduction of a state variable and care should be taken to reset it correctly.

https://github.com/user-attachments/assets/d270a0d9-da0d-4bb9-bba5-7c400e6a42a0

PR #115, opened 2024-10-01
fix: RangeSlider fixes

Letushev:
Fixes after review https://github.com/codux-templates/e-commerce-reclaim/pull/114


alisey (approved):
To make this component reusable in other templates, we should avoid using application-specific classes or variables in the component itself (`paragraph3`, `--primary5`). Instead, the component should accept a class name.

It sucks that customization requires using ugly selectors like .mySlider .thumbInput:focus-visible::-webkit-slider-thumb, so for basic customization, we could introduce variables:

```
--track-color
--thumb-color
--thumb-focus-ring-color
```

PR #117, opened 2024-10-02
refactor: improve RangeSlider customization

Letushev:
https://github.com/codux-templates/e-commerce-reclaim/pull/115#pullrequestreview-2340183264

### DIFF
--- a/src/components/range-slider/range-slider.tsx
+++ b/src/components/range-slider/range-slider.tsx
@@ -72,7 +72,7 @@ export const RangeSlider = ({
     };
 
     return (
-        <div className={classNames('rangeSlider', className)}>
+        <div className={classNames(styles.root, className)}>
             {/* The slider is implemented using three native <input type="range" />
                 elements stacked on top of each other.
             */}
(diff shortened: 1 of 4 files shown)


src/components/range-slider/range-slider.tsx
```diff
@@ -70,7 +72,7 @@ export const RangeSlider = ({
     };
 
     return (
-        <div>
+        <div className={classNames('rangeSlider', className)}>
```
alisey, line 75: I meant to pass `rangeSlider` to the instance. So this entire folder could be copied into a different project with no changes.

Ideally the component would come with some default hardcoded colors:
```
// range-slider.module.scss
.root {
  --thumb-color: #000;
}
```

But this would make overriding default styling tricky, since `common.scss` appears in the document earlier than `range-slider.module.scss`. But we could use `:where()` to control specificity:

https://jsfiddle.net/o94a8pn0/

Letushev: Done.
Plus found this bug in Codux. It says value is overriden when it is not.

Screenshot 2024-10-02 at 15 38 27

alisey: I guess because the specificity calculator doesn't understand `:where`.

alisey (approved):

PR #118, opened 2024-10-02
feat: filter products by price

Letushev:
https://github.com/user-attachments/assets/d8745f11-6529-4bc0-92c7-0b4fe8332b16

Notes:
- I am using `useSearchParams` to store filters in URL. Setting the search params causes a navigation and restarts route's loader.
- I hope `getProductPriceBounds` is temporary and we will get the API endpoint. 
- I'll ask about adding a loading indication next week


app/routes/products.$categorySlug/route.tsx
```diff
@@ -62,72 +81,148 @@ export const handle: RouteHandle<typeof loader> = {
...
+                    title="No products here yet..."
+                    subtitle="In the meantime, you can choose a different category to continue shopping."
+                />
+            );
+        }
+
+        if (someFiltersApplied && categoryProducts.items.length === 0) {
+            return (
+                <EmptyProductsCategory
+                    title="We couldn't find any matches"
```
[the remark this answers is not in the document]
Letushev: I copied messages from the wix website.

src/components/product-filters/product-filters.tsx
```diff
@@ -0,0 +1,65 @@
...
+export const ProductFilters = ({
+    appliedFilters,
+    onFiltersChange,
+    lowestPrice,
+    highestPrice,
+    formatPrice,
+}: ProductFiltersProps) => {
+    const navigation = useNavigation();
+
+    const [filters, setFilters] = useState(appliedFilters);
```
alisey, line 24: I don't quite understand this. If we're always updating state based on `appliedFilters`, then why do we need the internal state variable? Is it because `appliedFilters` updates with a delay?

src/components/product-filters/product-filters.tsx
```diff
@@ -0,0 +1,65 @@
...
+    useEffect(() => {
+        if (navigation.state !== 'loading') {
+            setFilters(appliedFilters);
+        }
+    }, [navigation.state, appliedFilters]);
+
+    return (
+        <Accordion
+            small
+            items={[
```
alisey, line 42: Unrelated to the PR, but I find providing items as children more intuitive / maybe easier to work with in Codux?

https://www.radix-ui.com/primitives/docs/components/accordion

Unfortunately their accordion doesn't change overflow to visible when expanded. (The MUI accordion does.)

src/api/ecom-api.ts
```diff
@@ -257,6 +267,29 @@ function createApi(): EcomAPI {
...
+                    .collection;
+                if (!category) throw new Error('Category not found');
+
+                const query = wixClient.products
+                    .queryProducts()
+                    .hasSome('collectionIds', [category._id]);
+
+                const [ascendingPrice, descendingPrice] = await Promise.all([
+                    await query.ascending('price').limit(1).find(),
+                    await query.descending('price').limit(1).find(),
```
alisey, line 282: Nice.

src/components/product-filters/product-filters.tsx
Letushev: > Is it because appliedFilters updates with a delay?

Yes. URL search parmas are not updated immediately. `setSearchParams` causes remix navigation, loader fetches new data and only after that search params are updated.

app/routes/products.$categorySlug/route.tsx
```diff
@@ -62,72 +81,148 @@ export const handle: RouteHandle<typeof loader> = {
...
+        useLoaderData<typeof loader>();
+
     const breadcrumbs = useBreadcrumbs();
 
+    const { filters, someFiltersApplied, applyFilters, clearFilters, clearAllFilters } =
+        useProductFilters();
+
+    const currency = categoryProducts.items[0]?.priceData?.currency;
+
+    const priceFormatter = useMemo(
```
alisey, line 94: How about adding a general-purpose utility `formatPrice(price, currency)` that can be imported directly by any component that wants to format the price?

app/routes/products.$categorySlug/use-product-filters.ts
```diff
@@ -0,0 +1,73 @@
+import { useCallback, useMemo } from 'react';
+import { useSearchParams } from '@remix-run/react';
+import { IProductFilters, ProductFilter } from '~/api/types';
+
+export function useProductFilters() {
```
alisey, line 5: If this can be used in other templates with no changes, let's move this to the `api` folder.

src/components/product-filters/product-filters.tsx
alisey: Let's add something like:

```ts
// Allows updating the UI optimistically while Remix navigates to the URL
// with updated parameters.
const [filters, setFilters] = useState(appliedFilters);
```

It's possible that after changing filters in the UI and older version of `appliedFilters` will arrive and temporarily undo the changes, but I don't have any suggestions how to prevent that without keeping track of versions.

src/components/product-filters/product-filters.tsx
Letushev: I'll take a look if I can do it with Radix, because I want to use Radix select for product variants.

alisey (approved):

PR #119, opened 2024-10-02
feat: allow accordion item content to overflow when expanded

Letushev:
I'm using `Accordion` in product filters. A range slider highlights a focused thumb with a box-shadow. Because accordion item has `overflow-y: auto` the overflowed box-shadow is cut off.

Screenshot 2024-10-02 at 18 17 35
<br /><br />

Using `transition-delay` and `transition-behavior: allow-discrete`  I reset the overflow after transiton ends.

### DIFF
--- a/app/routes/product-details.$productSlug/route.tsx
+++ b/app/routes/product-details.$productSlug/route.tsx
@@ -144,6 +144,7 @@ export default function ProductDetailsPage() {
                                     title: section.title!,
                                     content: section.description ? (
                                         <div
+                                            className={styles.additionalInfoDescription}
                                             dangerouslySetInnerHTML={{
                                                 __html: section.description,
                                             }}
--- a/src/components/accordion/accordion.tsx
+++ b/src/components/accordion/accordion.tsx
@@ -1,8 +1,9 @@
-import { useState } from 'react';
-import styles from './accordion.module.scss';
-import { MinusIcon, PlusIcon } from '../icons';
-import classNames from 'classnames';
+import { useRef, useState } from 'react';
+import { AnimatePresence, motion } from 'framer-motion';
 import { getClickableElementAttributes } from '~/utils';
+import { MinusIcon, PlusIcon } from '../icons';
+
+import styles from './accordion.module.scss';
 
 interface AccordionItem {
     title: string;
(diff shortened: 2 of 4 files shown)


alisey:
> I tried using transition-behavior but I didn't succeed.

`transition-delay` to the rescue https://jsfiddle.net/1Lkfo3sc/1/

---

For a JavaScript solution look at this:
https://github.com/codux-templates/e-commerce-reclaim/blob/1fe7f08bb182cf99cb211c57688f37610be84a82/src/components/visual-effects/reveal.tsx#L28-L29

src/components/accordion/accordion.tsx
```diff
@@ -22,32 +23,64 @@ export const Accordion = ({ items, className }: AccordionProps) => {
...
+
+interface AccordionItemProps extends AccordionItem {
+    isOpen: boolean;
+    onToggle: () => void;
+}
+
+const AccordionItem = ({ isOpen, onToggle, title, content }: AccordionItemProps) => {
+    const contentRef = useRef<HTMLDivElement | null>(null);
+
+    const isOpenRef = useRef(isOpen);
```
alisey, line 46: It doesn't look like you need `isOpenRef`, since it only changes if the prop changes, and `handleAnimationComplete` will be regenerated in that case.

Or does `motion.div` ignore changes to the `onAnimationComplete` prop?

Letushev: Yes, `motion.div` ignores the change so if I use `isOpen` it's not up to date.

alisey: 🥲👍

app/routes/product-details.$productSlug/route.tsx
```diff
@@ -144,6 +144,7 @@ export default function ProductDetailsPage() {
                                     title: section.title!,
                                     content: section.description ? (
                                         <div
+                                            className={styles.additionalInfoDescription}
```
alisey, line 147: Is this an additional design tweak, or is it somehow related to `overflow: visible`?

alisey (approved):

alisey:
> But made it without interpolate-size since it's support is far from good.

Yes, I've used `interpolate-size` just for the sake of simplicity in the demo. It's not widely supported yet.

app/routes/product-details.$productSlug/route.tsx
Letushev: Previously, content and padding were animated separately. It looked smooth but was still noticable because the timing-funciton is not linear. Padding expands from 0 to 25px, and content expands from 0 to auto.
I made that padding part of the item content, so that everything is animated together.

alisey: This makes default styling of accordion a bit sad:
image

Now padding needs to be added externally every time the component is used.
How about adding another wrapper or a spacer div into the component itself?

PR #121, opened 2024-10-03
fix: add wrapper to set accordion content padding

Letushev:


src/components/accordion/accordion.tsx
```diff
@@ -44,7 +44,9 @@ export const Accordion = ({ items, className }: AccordionProps) => {
                                 [styles.expanded]: isOpen,
                             })}
                         >
-                            <div className={styles.contentInner}>{item.content}</div>
+                            <div className={styles.contentExpander}>
+                                <div className={styles.contentWrapper}>{item.content}</div>
```
alisey, line 48: `contentInner` made more sense to me than `contentWrapper`, because usually the hierarchy is: `.contentWrapper > .content`.

alisey (approved):

PR #123, opened 2024-10-04
feat: add Select component

Letushev:
Add `Select` component using Radix. We need this to select product variants.

Screenshot 2024-10-04 at 17 30 45


_codux/boards/select/select.board.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+import styles from './select.board.module.scss';
+
+export default createBoard({
+    name: 'Select',
+    Board: () => {
+        const [value, setValue] = useState('');
+        return (
+            <div className={styles.container}>
+                <Select value={value} onValueChange={setValue} placeholder="Choose a pet">
```
tetianaman, line 13: I think it should be added to UI Kit "Components & Elements" board as well.

alisey (approved):

src/components/select/select.tsx
```diff
@@ -0,0 +1,43 @@
+import * as RadixSelect from '@radix-ui/react-select';
+import classNames from 'classnames';
+import { DropdownIcon } from '../icons';
+
+import styles from './select.module.scss';
+
+interface SelectProps {
```
alisey, line 7: Export the interfaces as well.

PR #131, opened 2024-10-08
fix: change the body background-color property to background shorthand

anatfenn:
to allow users to understand which variable is being used I changed the property of background-color to background so that the controller switches to the variable picker and users can understand what is their background color 
background controller doesn't show selected variable:
image

after change- variable name is revealed 
image


alisey:
Sorry, I forgot to tell you that in the template repos we don't created pull requests into the `main` branch, because that branch does directly to the users. Instead we open pull requests for the `development` branch.

alisey (approved):

PR #134, opened 2024-10-09
fix: price filter fixes

Letushev:
1. When `min` or `max ` of `<input type="range" />` is a floating number and the step is an integer, the slider looks weird:
Screenshot 2024-10-09 at 12 46 24

Plain HTML https://jsfiddle.net/v6xpecL0/

To overcome this I set `step="any"` to the price range slider.
Additionally, I round values on change, because:
1. It's not convinient to slide between decimals.
2. It's nicer to have integer in URL than `?maxPrice=11.1199717420213`
<hr />

2. Previously, applied price filter looked bad after changing only mimumum or only maximum:

Screenshot 2024-10-09 at 12 53 09
<br /><br />

The fix is to show the lowest/highest value if it's not applied explicitly.

Screenshot 2024-10-09 at 12 56 24

### DIFF
--- a/src/components/applied-product-filters/applied-product-filters.tsx
+++ b/src/components/applied-product-filters/applied-product-filters.tsx
@@ -11,7 +11,10 @@ interface AppliedProductFiltersProps {
     onClearFilters: (filters: ProductFilter[]) => void;
     onClearAllFilters: () => void;
     currency: string;
-    defaultPriceFilters: { minPrice: number; maxPrice: number };
+    // Min and max prices in the current category.
+    // Used to replace missing bounds ("$5 - ?" or "? - $25") when only one filter bound is set.
+    minPriceInCategory: number;
+    maxPriceInCategory: number;
     className?: string;
 }
 
@@ -20,7 +23,8 @@ export const AppliedProductFilters = ({
     onClearFilters,
     onClearAllFilters,
     currency,
-    defaultPriceFilters,
+    minPriceInCategory,
+    maxPriceInCategory,
     className,
 }: AppliedProductFiltersProps) => {
     const { minPrice, maxPrice } = appliedFilters;
@@ -31,12 +35,12 @@ export const AppliedProductFilters = ({
         } else {
             return (
                 <span>
-                    {formatPrice(minPrice ?? defaultPriceFilters.minPrice, currency)}&ndash;
-                    {formatPrice(maxPrice ?? defaultPriceFilters.maxPrice, currency)}
+                    {formatPrice(minPrice ?? minPriceInCategory, currency)}&ndash;
+                    {formatPrice(maxPrice ?? maxPriceInCategory, currency)}
                 </span>
             );
         }
-    }, [minPrice, maxPrice, currency, defaultPriceFilters]);
+    }, [minPrice, maxPrice, currency, minPriceInCategory, maxPriceInCategory]);
 
     return (
         <div className={classNames(styles.root, className)}>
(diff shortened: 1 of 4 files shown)


alisey:
> Previously, applied price filter looked bad after changing only mimumum or only maximum

The screenshots before and after look identical to me.

alisey:
Ah, you mean this
image

src/components/applied-product-filters/applied-product-filters.tsx
```diff
@@ -11,6 +11,7 @@ interface AppliedProductFiltersProps {
     onClearFilters: (filters: ProductFilter[]) => void;
     onClearAllFilters: () => void;
     currency: string;
+    priceBounds: { lowest: number; highest: number };
```
alisey, line 14: The component receives both `appliedFilters: { minPrice, maxPrice }` and `priceBounds: { lowest; highest }`. I don't understand the difference between them.

src/components/product-filters/product-filters.tsx
```diff
@@ -53,13 +53,18 @@ export const ProductFilters = ({
                     content: (
                         <RangeSlider
                             className="rangeSlider"
+                            step="any"
                             startValue={filters.minPrice ?? lowestPrice}
                             endValue={filters.maxPrice ?? highestPrice}
                             onStartValueChange={(value) => {
-                                handleFiltersChange({ minPrice: value });
+                                handleFiltersChange({
+                                    minPrice: value === lowestPrice ? value : Math.round(value),
```
alisey, line 61: I think it should be `Math.floor` for the lower bound and `Math.ceil` for the upper bound to ensure we're not excluding results in the filter's range.

src/components/applied-product-filters/applied-product-filters.tsx
Letushev: `appliedFilters` are filters from URL. I use lowest/highest from `priceBounds` as default values if one of them is missing. Is it more clear if I rename `priceBounds` to `defaultPriceFilters`?

alisey: Now I'm spoilt by the knowledge and it's difficult to judge, but maybe:
```ts
// Min and max prices in filtered results. Used to replace missing
// bounds ("$5 - ?" or "? - $25") when only one filter bound is set.
minPriceFound: number;
maxPriceFound: number;
```

I think this would have an issue when no products are found for the given filters?

Letushev: > I think this would have an issue when no products are found for the given filters?

No (if I get the question correctly). Price bounds don't depend on what filters are currently applied.

alisey: But they do, screenshots from this branch, in both cases the price filter is `minPrice=10`:

Screenshot 2024-10-09 at 19 44 46
Screenshot 2024-10-09 at 19 45 00

Letushev: Ah, now it seems to me that the first version was good enough. What if I return as it was and add a comment that you suggested?
```tsx
/**
 * The highest and lowest product prices in the current category.
 * Used to replace missing bounds ("$5 - ?" or "? - $25") when only one filter bound is set.
 */
priceBounds: { lowest: number; highest: number };
```

Letushev: > But they do, screenshots from this branch, in both cases the price filter is minPrice=10:

I am confused. Those are different categories on screenshots.

Letushev: @alisey I've pushed new chages.

alisey (approved):
cool cool cool

PR #135, opened 2024-10-09
feat: product sorting

Letushev:
Allow to sort category products.
I didn't add "Recommended" because API doens't support this option yet. The dafault is "Newest".

Screenshot 2024-10-10 at 13 12 53


src/components/product-sorting-select/product-sorting-select.module.scss
```diff
@@ -0,0 +1,31 @@
+.select {
```
Letushev, line 1: This select looks different from the select to choose product variants, which I consider has the default select styles.

app/routes/products.$categorySlug/route.module.scss
```diff
@@ -98,8 +105,19 @@
...
+    }
+
+    .productsCount {
+        align-self: start;
+    }
+}
+
 @media (max-width: $mobile-width) {
-    .navigation {
+    .sidebar {
```
Letushev, line 120: Not related to PR, but it's a small thing I forgot to change when added filters.

src/api/product-filters.ts
```diff
@@ -62,11 +62,10 @@ export function parseProductFiltersFromUrlSearchParams(
     };
 }
 
-export function stringifyProductFiltersToUrlSearchParams({
-    minPrice,
-    maxPrice,
-}: IProductFilters): URLSearchParams {
-    const searchParams = new URLSearchParams();
+export function setProductFiltersToUrlSearchParams(
```
alisey, line 65: It's ambiguous whether it's `setProductFilters` or `setUrlSearchParams`.

One alternative:
```
convertProductFiltersToUrlParams(filters): URLSearchParams
convertUrlParamsToProductFilters(params): IProductFilters

mergeUrlParams(a: URLSearchParams, b: URLSearchParams): URLSearchParams
```

src/utils/use-optimistic-search-params-state.ts
```diff
@@ -0,0 +1,31 @@
+import { useNavigation } from '@remix-run/react';
+import { useCallback, useEffect, useState } from 'react';
+
+/**
+ * Allows updating the UI optimistically while Remix navigates to the URL with updated parameters.
+ */
+export function useOptimisticSearchParamsState<S>(
+    stateFromSearchParams: S,
+    setStateToSearchParams: (state: S) => void,
```
alisey, line 9: Same ambiguity as above. I'm giving it state, so I guess this callback should update search params? But if the hook is only for updating UI optimistically, why is it changing search params?

I'm quite confused about what this hook is responsible for.

src/components/product-sorting-select/product-sorting-select.module.scss
```diff
@@ -0,0 +1,31 @@
...
+    font: var(--paragraph3);
+}
+
+.selectDropdown {
+    padding: 12px 0;
+    border: 1px solid rgb(23, 17, 13, 0.4);
+    background: var(--primary1);
+    box-shadow:
+        0 4px 8px 0 rgba(0, 0, 0, 0.1),
+        0 0 4px 0 rgba(0, 0, 0, 0.1);
```
tetianaman, line 15: Lets maybe use `rgba` having 4th param?

src/components/product-sorting-select/product-sorting-select.module.scss
```diff
@@ -0,0 +1,31 @@
...
+}
+
+.selectItem {
+    height: 38px;
+    padding: 8px 24px;
+    font: var(--paragraph3);
+}
+
+.selectItem[data-state='checked'] {
+    background: rgb(23, 17, 13, 0.12);
```
tetianaman, line 25: the same here and below

src/components/product-sorting-select/product-sorting-select.tsx
```diff
@@ -0,0 +1,58 @@
...
+    { value: ProductSortBy.priceAsc, label: 'Price (low to high)' },
+    { value: ProductSortBy.priceDesc, label: 'Price (high to low)' },
+    { value: ProductSortBy.nameAsc, label: 'Name A-Z' },
+    { value: ProductSortBy.nameDesc, label: 'Name Z-A' },
+];
+
+export const ProductSortingSelect = () => {
+    const [searchParams, setSearchParams] = useSearchParams();
+
+    const appliedSortBy = useMemo(
```
tetianaman, line 21: Would it be a good idea to move sorting to separate hook to make it possible to use in other projects?

tetianaman: but it is already a component. Ok.

src/utils/use-optimistic-search-params-state.ts
Letushev: > setStateToSearchParams - same ambiguity as above, I'm giving this function state, so I guess it should update search params? But if the hook is only for updating UI optimistically, why is it changing search params?

Not sure I get it.

The hook is responsible for providing react state that we can update immediately and sync it with the state in URL search params.

src/components/product-sorting-select/product-sorting-select.module.scss
alisey: @tetianaman why? rgba is legacy syntax that is not recommended to use anymore.

src/utils/use-optimistic-search-params-state.ts
alisey: The part "keep it in sync it with the state in URL search params" is not clear to me.

I'm looking at the params, at the return value `return [state, handleChange]`, at the function description, and things don't connect in my head.

src/components/product-sorting-select/product-sorting-select.module.scss
Letushev: From spec I understood that comma separated syntax is legacy, not `rgba` function.
But I don't see any difference between modern rgb and rgba syntaxes.

I'll change it to `rgb(23 17 13 / 0.4)`.

alisey: image

Letushev: I keep using `rgba()`, becuase Codux color picker doesn't support modern rgb syntax.

tetianaman (approved):

src/components/select/select.tsx
```diff
@@ -4,24 +4,38 @@ import { DropdownIcon } from '../icons';
...
-    value: string;
-    onValueChange: (value: string) => void;
+export interface SelectProps<V extends string> {
+    value: V;
+    onValueChange: (value: V) => void;
     placeholder?: string;
     children: React.ReactNode;
+    className?: string;
+    dropdownClassName?: string;
+    customSelectedValue?: React.ReactNode;
```
alisey, line 14: I don't understand the difference between this and placeholder.

alisey: Oh, is it like `renderValue`?

Letushev: Yes

Letushev: Updated

alisey (approved):

PR #138, opened 2024-10-10
feat: color select

Letushev:
Screenshot 2024-10-10 at 17 06 32

`options` is not `string[]` (unfortunately) because for products it's allowed to add colors with the same color value.
What must be unique is the name given to color. So we will use name as the color id.

### DIFF
--- a/_codux/boards/color-select.board.tsx
+++ b/_codux/boards/color-select.board.tsx
@@ -1,0 +1,22 @@
+import { useState } from 'react';
+import { createBoard } from '@wixc3/react-board';
+import { ColorSelect, ColorSelectOption } from '~/components/color-select/color-select';
+
+const options: ColorSelectOption[] = [
+    { id: 'color1', color: '#00a400' },
+    { id: 'color2', color: 'rgb(214, 122, 127)' },
+    { id: 'color3', color: 'hsl(30deg 82% 43%)' },
+    { id: 'color4', color: 'rebeccapurple' },
+];
+
+export default createBoard({
+    name: 'Color Select',
+    Board: () => {
+        const [color, setColor] = useState('');
+        return <ColorSelect options={options} selectedId={color} onChange={setColor} />;
+    },
+    environmentProps: {
+        windowWidth: 400,
+        windowHeight: 100,
+    },
+});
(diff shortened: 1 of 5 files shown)


_codux/boards/color-select.board.tsx
```diff
@@ -0,0 +1,22 @@
+import { useState } from 'react';
+import { createBoard } from '@wixc3/react-board';
+import { ColorSelect, ColorSelectOption } from '~/components/color-select/color-select';
+
+const options: ColorSelectOption[] = [
+    { id: 'color1', color: '#00a400' },
+    { id: 'color2', color: 'rgb(214, 122, 127)' },
+    { id: 'color3', color: 'hsl(30deg 82% 43%)' },
+    { id: 'color4', color: 'rebeccapurple' },
```
alisey, line 9: Let's include `black` and `white`, and add padding to the board.

src/components/color-select/color-select.tsx
```diff
@@ -0,0 +1,36 @@
+import classNames from 'classnames';
+
+import styles from './color-select.module.scss';
+
+export interface ColorSelectOption {
+    id: string;
+    color: string;
+}
+
+export interface ColorSelectProps {
```
alisey, line 10: Add support for className.

src/components/color-select/color-select.tsx
```diff
@@ -0,0 +1,36 @@
...
+    options: ColorSelectOption[];
+    selectedId: string;
+    onChange: (id: string) => void;
+    hasError?: boolean;
```
alisey, line 14: What `hasError` is used for? To highlight that this is a required field?

src/components/color-select/color-select.module.scss
```diff
@@ -0,0 +1,41 @@
+.root {
+    display: flex;
+    gap: 8px;
+}
+
+.option {
+    height: 20px;
+    width: 20px;
+    border: 1px solid var(--primary3);
```
alisey, line 9: Let's do what we did for the slider component or this could be reused with no changes in the other template.

src/components/color-select/color-select.tsx
Letushev: Yes
Screenshot 2024-10-11 at 10 48 38

alisey (approved):

PR #141, opened 2024-10-11
feat: add logic for cart-page

tetianaman:
Cart page works now. 
No updating states over cart item and cart summary in this PR 

https://github.com/user-attachments/assets/216a9d2c-bd70-4499-ba21-4439b4ff1219

Data synchronisation between tabs:

https://github.com/user-attachments/assets/4d48cb64-d7ed-4471-8a7f-6e1432d425f5


tetianaman:
@yurii-ve 
As for the first video,
pay attention to the delay between changing an item and the total prices being recalculated.
IMO, it looks okay, but having slow internet connection could be a problem. 2-3 sec could be stressful for buyer.
Anyway, during cart or price updating we need to disable checkout button.

tetianaman:
@yurii-ve 
I've just tried to go forward and implement updating states for cart items and cart summary. 
I've tried to put loader over the cart item which quantity was changed. I want to remove loader after price summary is recalculated.
What I stuck with is that `cartTotals` is being updated automatically because `cart` was updated. I don't need to trigger `cartTotalsMutate` on item change success. Therefore the updating cart item doesn't know when "its" `cartTotals` update is finished. See comments in the code below.

_useCart hook_
```
    const [updatingCartItemIds, setUpdatingCartItemIds] = useState<string[]>([]);
    
    const updateItemQuantity = ({ id, quantity }: { id: string; quantity: number }) => {
        setUpdatingCartItemIds((prev) => [...prev, id]);
        triggerUpdateItemQuantity(
            { id, quantity },
            {
                onSuccess: () =>
                    // how to know that cartTotals mutation after this quantity change is done?
                    // no await cartTotalsMutate() here
                    setUpdatingCartItemIds((prev) => prev.filter((itemId) => itemId !== id)),
            },
        );
    };
```

tetianaman:
Btw, for now we have
home or category page onload: one _current_, 4 _estimate-totals_ requests
cart page onload: one _current_, 7 _estimate-totals_ requests

src/components/cart/cart-view/cart-view.tsx
```diff
@@ -75,6 +77,12 @@ export const CartView = ({
                         >
                             Checkout
                         </button>
+                        <button
```
Letushev, line 80: Make it a link

src/api/api-hooks.ts
```diff
@@ -6,28 +6,40 @@ import { useEcomAPI } from './ecom-api-context-provider';
...
+            if (response.status === 'failure') {
+                throw response.error;
+            }
 
-        return response.body;
-    });
+            return response.body;
+        },
+        {
+            revalidateOnFocus: true,
```
Letushev, line 20: I think we should set it for global SWR config.

app/routes/cart/route.tsx
```diff
@@ -1,55 +1,63 @@
...
                     <div className={styles.summaryRow}>
                         <span>Subtotal</span>
-                        <span>$5.50</span>
+                        <span>{cartTotals?.priceSummary?.subtotal?.formattedConvertedAmount}</span>
                     </div>
                     <div className={styles.summaryRow}>
                         <span>Delivery</span>
-                        <span>FREE</span>
+                        <span>
+                            {cartTotals?.priceSummary?.shipping?.amount === '0.00'
```
Letushev, line 57: I think `Number(priceSummary?.shipping?.amount) === 0` would be more reliable. For example, api can suddently change formatting to one number of decimal places.

app/routes/cart/route.tsx
```diff
@@ -1,55 +1,63 @@
...
-const noop = () => {};
+    if (!cartData) return null;
+
+    if (!cartData.lineItems.length)
+        return (
+            <div className={styles.cart}>
+                <h1 className={styles.cartHeader}>My cart</h1>
+                <div className={styles.emptyCartMessage}>
+                    <div>Cart is empty</div>
+                    <NavLink to={ROUTES.home.to()} className={styles.continueBrowsingLink}>
```
Letushev, line 22: `Link` instead of `NavLink`

app/routes/cart/route.tsx
```diff
@@ -1,55 +1,63 @@
 import { CartItem } from '~/components/cart/cart-item/cart-item';
-import { cart } from '@wix/ecom';
 import classNames from 'classnames';
 import { LockIcon } from '~/components/icons';
+import { ROUTES } from '~/router/config';
+import { NavLink } from '@remix-run/react';
 
 import styles from './route.module.scss';
+import { useCart } from '~/hooks/use-cart';
+import { findLineItemPriceBreakdown } from '~/api/cart-helpers';
```
Letushev, line 9: Make styles import last and separated.

src/components/cart/cart-item/cart-item.tsx
```diff
@@ -29,6 +29,10 @@ export const CartItem = ({
 
     const [quantity, setQuantity] = useState(item.quantity!);
 
+    useEffect(() => {
+        setQuantity(item.quantity!);
+    }, [item.quantity]);
```
Letushev, line 34: I'm afraid this can override the local `quantity`. For example, you update quanity from 1 to 2. While request is in progress you change it from 2 to 3. Then the request has finished and it overrides local state from 3 to 2. I think to prevent this you can do:
```js
if (!isUpdating) {
  setQuantity(item.quantity!);
}
```

Letushev:
> cart page onload: one current, 7 estimate-totals requests

Why so many?

Letushev:
> Therefore the updating cart item doesn't know when "its" cartTotals update is finished.

You have an SWR hook that fetches cart totals. It gives you `isLoading`. Can't it be as simple as:
```
const isLoading = isCartItemUpdating || isCartTotalsLoading;
```

tetianaman:
> > cart page onload: one current, 7 estimate-totals requests
> 
> Why so many?

I guess useEffect works so.
We have three usages of  `const { data: cart } = useCartData();` (Header, Cart, CartPage) and one usage of `useCartTotals();` in `UseCart` hook.

Then we have in `UseCartTotals` hook: 
```
    useEffect(() => {
        cartTotals.mutate();
    }, [data]);
```
So for each usage `useEffect` executes twice. It's 6 + one swr request for cart-totals. 7.

src/components/cart/cart-item/cart-item.tsx
tetianaman: I'd rather disable changing quantity the same as checkout button on data updating. Because in your case we can loose data also (for e.g. if user changes quantity in Cart panel and cart page is open, it should update quantity on the cart-page as well)

Letushev: Disabling will not allow users to quickly change quantity by several units. They will have to do it slowly, one at a time.

src/components/cart/cart-view/cart-view.tsx
tetianaman: I would stay with button, because I need to close right panel. 
```
  const onViewCart = () => {
    setIsOpen(false);
    navigate(ROUTES.cart.to());
  };
```
Look not nice:
```
<Link to={ROUTES.cart.to()}
      onClick={onViewCart} 
>
```

Letushev: I don't see a problem with `onClick`. Looks ok to me.

src/components/cart/cart-view/cart-view.tsx
```diff
@@ -75,6 +79,13 @@ export const CartView = ({
                         >
                             Checkout
                         </button>
+                        <Link
```
yurii-ve, line 82: I know that @Letushev asked to use Link, but I don't agree with this change. It's a view component, and `onViewCart` should decide if it's navigation or maybe something else. On the practical matter - now to create a board for the card view it's required to wrap it in some Remix (or Router) context.

yurii-ve: In general I AGREE that links should be links, but one of the objectives of creating `CartView` component was to be able to create simple board without any mocks and context. The remix/react-router `Link` won't allow to do that.

tetianaman: I will add Cart board in separate PR

Letushev: But we just need to wrap board component with `ComponentWrapper`. What's so bad about that?

tetianaman: We can also keep `Link` and pass `cartPageUrl` prop.

tetianaman: > What's so bad about that?
It shows that our view component is not "good enough" view :)

yurii-ve: > But we just need to wrap board component with ComponentWrapper. What's so bad about that?

`ComponentWrapper` wraps board component in `EcomApiProvider` and creates Remix stub. As a result board elements tree is not that straightforward and it's just adds unnecessary complexity for the user in my opinion.
So if choosing between simpler board and having this single link as a link I prefer simpler board.

Letushev (approved):

PR #143, opened 2024-10-14
fix: move buttons styles out of `_codux` directory

TheAlmightyCrumb:
These were disabled before because imports from `_codux` throw a 'development-only code' error while processing the snippet.

image


tetianaman:
The fix is in[ this PR](https://github.com/codux-templates/e-commerce-reclaim/pull/145)

PR #144, opened 2024-10-14
refactor: product filters in URL search params

Letushev:
Add `useSearchParamsOptimistic` hook to simplify working with the state that lives in URL search params.
This will come in handy for other states in URL, such as product sorting.

### DIFF
--- a/src/utils/use-search-params-state.ts
+++ b/src/utils/use-search-params-state.ts
@@ -1,0 +1,40 @@
+import { useCallback, useEffect, useState } from 'react';
+import type { NavigateOptions } from 'react-router';
+import { useNavigation, useSearchParams } from '@remix-run/react';
+import { mergeUrlSearchParams } from './index';
+
+export interface SearchParamsStateConverter<S> {
+    fromSearchParams: (params: URLSearchParams) => S;
+    toSearchParams: (state: S) => URLSearchParams;
+}
+
+export function useSearchParamsState<S>(
+    converter: SearchParamsStateConverter<S>,
+    navigateOptions: NavigateOptions = { preventScrollReset: true },
+) {
+    const navigation = useNavigation();
+    const [searchParams, setSearchParams] = useSearchParams();
+
+    // Allows updating the UI optimistically
+    // while Remix navigates to the URL with updated parameters.
+    const [state, setState] = useState(() => converter.fromSearchParams(searchParams));
+
+    const handleChange = useCallback(
+        (newState: S) => {
+            setState(newState);
+            setSearchParams((prevParams) => {
+                return mergeUrlSearchParams(prevParams, converter.toSearchParams(newState));
+            }, navigateOptions);
+        },
+        [setSearchParams, converter, navigateOptions],
+    );
+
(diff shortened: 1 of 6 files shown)


src/utils/use-search-params-state.ts
```diff
@@ -0,0 +1,48 @@
...
+import { mergeUrlSearchParams } from './index';
+
+export interface SearchParamsStateConverter<S> {
+    fromSearchParams: (params: URLSearchParams) => S;
+    toSearchParams: (state: S) => URLSearchParams;
+}
+
+/**
+ * Helper to manage the state that lives in URL search params.
+ * Search params are not updated immediately
```
alisey, line 13: Not updated immediately by what? By the hook?

Letushev: By Remix

src/utils/use-search-params-state.ts
```diff
@@ -0,0 +1,48 @@
...
+ * because changing them causes a Remix navigation and loaders reloading.
+ * This hooks provides a local React state to apply changes to URL optimistically
+ * and ensures it's up to date with the values in URL.
+ *
+ * @param converter Converts specific state from search params and vice versa.
+ * @param navigateOptions Router navigation options when changing search params.
+ */
+export function useSearchParamsState<S>(
```
alisey, line 21: The first thing that comes to mind:
```ts
const [searchParams, setSearchParams] = useSearchParamsOptimistic();

const filters = productFiltersFromSearchParams(searchParams);

const updateFilters = (newFilters) => {
    const newParams = searchParamsFromProductFilters(newFilters);
    setSearchParams(params => mergeSearchParams(params, newParams));
};
```

This is a bit more verbose, but I think it's an easier concept to grasp. "It's like useSearchParams(), but updated optimistically". This hook can also can be used when no convertor is necessary, e.g. we just want to read/write a single string param.

This also sidesteps the tricky question of naming: `useSearchParamsState` reads like "it's the state of search params". And then I start thinking, isn't it what `useSearchParams()` is?

And it also sidesteps the question of whether new and old parameters are merged or replaced.

But I'll leave this up to you. If you still think that `useSearchParamsState` is a better option, let's go with that.

src/utils/use-search-params-state.ts
alisey: It's not clear from the comment.

Also the following "because": I'm not sure how one leads to another. Remix could `history.pushState` immediately, it doesn't do that, but not because of navigation.

src/utils/use-search-params-state.ts
alisey: When I'm thinking about it like this:

```ts
const [filters, setFilters] = useStateSyncedToSearchParams({ serializeState, parseState });
```

It is pretty intuitive, except the part with how merging works.

alisey (approved):

src/utils/use-search-params-state.ts
Letushev: I've tried `useSearchParamsOptimistic`. The implementation is more straightforward and it looks easier to grasp. Thank you for the idea, take a look one more time please.

alisey (approved):

PR #147, opened 2024-10-14
feat: add updating state to CartItem and Cart

tetianaman:
I'd like the loaders over cart item and cart summary start and finish simultaneously. 
It cannot be implemented easily due to automatic `CartTotals` update on `Cart` change ([see `useCartTotals` api-hook](https://github.com/codux-templates/e-commerce-reclaim/pull/147/files#diff-c336644a0a24a2c16d2910978e0f6c9d7d688800c5aa340c20ab90c51e6a41f3R33)). To keep loader over cart item, the item should know when related to it `CartTotals` update finished. 

I suggest to accept this PR solution so far: loader over cart item finished a bit earlier than prices updated.

https://github.com/user-attachments/assets/cc2b5887-c392-45ae-9063-8e26e08f412c


src/components/cart/cart-view/cart-view.tsx
```diff
@@ -14,6 +14,8 @@ export interface CartViewProps {
     onViewCart: () => void;
     onItemQuantityChange: (args: { id: string; quantity: number }) => void;
     onItemRemove: (id: string) => void;
+    isCartItemUpdating?: (id: string) => boolean;
+    isCartUpdating?: boolean;
```
yurii-ve, line 18: just `isUpdating`? or maybe even `isCheckoutAllowed`?

src/api/api-hooks.ts
```diff
@@ -19,6 +19,7 @@ export const useCartData = () => {
 export const useCartTotals = () => {
     const ecomApi = useEcomAPI();
     const { data } = useCartData();
+    const [isUpdatingOnCartChange, setIsUpdatingOnCartChange] = useState(false);
```
yurii-ve, line 22: do we really care if it's loading only due to cart change? can't we use `cartTotals.isLoading`?

src/components/cart/cart-view/cart-view.tsx
```diff
@@ -14,6 +14,8 @@ export interface CartViewProps {
     onViewCart: () => void;
     onItemQuantityChange: (args: { id: string; quantity: number }) => void;
     onItemRemove: (id: string) => void;
+    isCartItemUpdating?: (id: string) => boolean;
```
yurii-ve, line 17: there are two props (`isCartItemUpdating` and `isCartUpdating`) named almost exactly the same but one of the is a function and another - boolean. Let's maybe rename this to `checkIfCartItemUpdating` or something in this manner.

another think to consider is just passing `updatingItemIds`.. I don't have strong arguments for that but just think that props that are primitives (or array of primitives) are easier to grasp than functions.

src/hooks/use-cart.ts
```diff
@@ -8,10 +9,27 @@ import { useEcomAPI } from '~/api/ecom-api-context-provider';
...
+        useCartTotals();
+    const cartTotals = cartTotalsData?.data;
+
+    const { trigger: triggerUpdateItemQuantity } = useUpdateCartItemQuantity();
+    const { trigger: triggerRemoveItem } = useRemoveItemFromCart();
+
+    const updateItemQuantity = async ({ id, quantity }: { id: string; quantity: number }) => {
+        setUpdatingCartItems((prev) => [...prev, id]);
+        await triggerUpdateItemQuantity({ id, quantity });
+        setUpdatingCartItems((prev) => prev.filter((itemId) => itemId !== id));
```
Letushev, line 25: It's super incovinient to do it like this. Can't we move hooks to the cart item component and use `updateCartItemQuantity.isLoading`?

app/routes/cart/route.module.scss
```diff
@@ -24,11 +24,20 @@
 }
 
 .summarySection {
+    position: relative;
     padding: 14px 0;
 }
 
-.summarySection:not(:last-child) {
-    border-bottom: 1px solid rgba(23, 17, 13, 0.2);
+.summarySection.updating {
```
yurii-ve, line 31: `updatingCover`?

src/api/api-hooks.ts
tetianaman: `cartTotals.isLoading` - this is the period when data is fetching for the first time
`cartTotals.isValidating` - this is the period when data is fetching to update the cache and it works also on 
_SWRConfig_
```
   revalidateOnFocus: true,
   revalidateOnReconnect: true,
   refreshInterval: 5 * 60_000, // 5 minutes
```

Using `isValidating` makes checkout button blinking (from normal to disable state) on each onFocus and it look strange.

app/routes/cart/route.tsx
```diff
@@ -46,7 +56,11 @@ export default function CartPage() {
             </div>
             <div className={styles.summary}>
                 <h1 className={styles.summaryHeader}>Order summary</h1>
-                <div className={styles.summarySection}>
+                <div
+                    className={classNames(styles.summarySection, {
+                        [styles.updating]: isCartTotalsUpdating,
```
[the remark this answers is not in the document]
tetianaman: I don't like it. Loading is when data is absent and we're waiting for it. Here we have data. It is updating.

src/api/api-hooks.ts
yurii-ve: we fetch cart totals each time user hovers on the `Checkout` button?

src/hooks/use-cart.ts
tetianaman: - The idea was to move cart logic into one hook to use easily between similar projects.
- `CartItem` is view component

src/api/api-hooks.ts
tetianaman: no, browser tab onFocus

src/hooks/use-cart.ts
Letushev: But if because of that we need to write something like that, I'm against it

tetianaman: I think it's a bit better.

```
    const updateItemQuantity = ({ id, quantity }: { id: string; quantity: number }) => {
        setUpdatingCartItems((prev) => [...prev, id]);
        triggerUpdateItemQuantity(
            { id, quantity },
            {
                onSuccess: () =>
                    setUpdatingCartItems((prev) => prev.filter((itemId) => itemId !== id)),
            },
        );
    };
```

src/api/api-hooks.ts
Letushev: The option I see is to remove that `useEffect` that mutates cart totals on cart change.
If cart totals depend on cart, then we should reflect in key https://swr.vercel.app/docs/conditional-fetching
For example, using cart updated date:
```tsx
const { data } = useCartData();
const cartTotals = useSwr('cart-totals' + data._updatedDate', fn)

cartTotals.isLoading; // true when cart totals refetch after cart change, not on revalidate
```

src/hooks/use-cart.ts
Letushev: It's the same

yurii-ve: @Letushev I'm with you :) but don't forget that it's a template for Codux.. and for Codux users it's good to have as many dumb components as possible so they are easy to work with in boards.

src/api/api-hooks.ts
tetianaman: @Letushev let me test it

src/hooks/use-cart.ts
yurii-ve: to be honest when I refactored the code in ecom-starter template (extracting view components) I wasn't aware of (and didn't think of) this updating items UI. Refactoring made sense then.. now I agree that it's more on the edge but still believe that having pure view components is highly beneficial in our context.

@alisey what do you think?

alisey: I don't want to get deep into this, but `setUpdatingCartItems` doesn't look too bad. "Here's a list of items currently being updated". Makes sense. If it's hidden inside the hook, and the user doesn't need to worry about, it's a win for me.

src/api/api-hooks.ts
tetianaman: @Letushev I've played with it a bit.
`() =>` `'cart-totals/${data._updatedDate}` - this generates many caches. We will need to clean old cache. 
`() => cart ? 'cart-totals' : null` - this doesn't react on cart change.

Letushev: Okay, your solution or `isValidating` is fine then

tetianaman: > maybe it's ok to have the button disabled in such case

Agree. Kept `isValidating`

app/routes/cart/route.module.scss
tetianaman: just `.loading`

app/routes/cart/route.tsx
```diff
@@ -7,9 +7,18 @@ import { useCart } from '~/hooks/use-cart';
...
 import styles from './route.module.scss';
+import { Spinner } from '~/components/spinner/spinner';
 
 export default function CartPage() {
-    const { cartData, cartTotals, checkout, removeItem, updateItemQuantity } = useCart();
+    const {
+        cartData,
+        cartTotals,
+        isCartTotalsUpdating,
+        updatingCartItems,
```
yurii-ve, line 17: `updatingCartItemIds`

src/hooks/use-cart.ts
```diff
@@ -1,17 +1,43 @@
...
-    const { trigger: removeItem } = useRemoveItemFromCart();
+    const { data: cartTotals, isValidating: isCartTotalsValidating } = useCartTotals();
+
+    const { trigger: triggerUpdateItemQuantity } = useUpdateCartItemQuantity();
+    const { trigger: triggerRemoveItem } = useRemoveItemFromCart();
+    const { trigger: triggerAddToCart, isMutating: isAddingToCart } = useAddToCart();
+
+    const updateItemQuantity = ({ id, quantity }: { id: string; quantity: number }) => {
+        setUpdatingCartItems((prev) => [...prev, id]);
+        triggerUpdateItemQuantity(
```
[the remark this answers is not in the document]
alisey: And even then, I don't know. I can send two requests at a small interval to update quantity of item A, once the first request completes, it will remove the item from the list.

If it's only used for the loader, then it's ok.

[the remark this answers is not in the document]
tetianaman: Oh, it's quite hard to make such two requests. We have debounce for fast quantity changes (300 ms), after that request is sent and cart item is disabled under the loader. And loader disappears because the item is removed from an array on request success or error.

As for failure, let me do it in separate PR:

Screenshot 2024-10-15 at 19 10 53

Screenshot 2024-10-15 at 19 10 33

src/hooks/use-cart.ts
```diff
@@ -26,8 +49,14 @@ export const useCart = () => {
...
         cartData,
         cartTotals,
+        updatingCartItemIds,
+
+        isAddingToCart,
+        isCartTotalsUpdating: updatingCartItemIds.length > 0 || isCartTotalsValidating,
+
         updateItemQuantity,
         removeItem,
+        addToCart,
```
yurii-ve, line 59: `addItem`?

yurii-ve (approved):

PR #150, opened 2024-10-15
feat: support product variants

Letushev:
Copy product variants functionality from https://github.com/codux-templates/e-commerce-remix
The only differences are in styles and components that we use.

PR includes:
- Allow to choose product options.
- Adjust product details depending on the selected options.
- Update API to handle adding product with options to the cart.

To be done:
- Display selected options in cart. I'll do this in a separate PR https://github.com/codux-templates/e-commerce-reclaim/pull/151
- Hidden and out of stock variants. I'll wait for https://github.com/codux-templates/e-commerce-remix/pull/89

### DIFF
--- a/app/routes/product-details.$productSlug/route.module.scss
+++ b/app/routes/product-details.$productSlug/route.module.scss
@@ -35,6 +35,13 @@
     line-height: 1.5;
 }
 
+.productOptions {
+    display: flex;
+    flex-direction: column;
+    gap: 16px;
+    margin-top: 14px;
+}
+
 .quantity {
     margin-top: 20px;
 }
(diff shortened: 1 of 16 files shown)


app/routes/product-details.$productSlug/route.module.scss
```diff
@@ -35,6 +35,13 @@
     line-height: 1.5;
 }
 
+.productOptions {
+    display: flex;
```
tetianaman, line 39: why `flex`? To avoid styles like `.myClass:not(:last-child)` ? What are other reasons?

tetianaman (approved):

app/routes/product-details.$productSlug/route.module.scss
Letushev: Just the simplest way to add gap between elements.

PR #156, opened 2024-10-17
feat: add error handling for the cart

tetianaman:
1. Toast and Toaster components have been added. The idea is that Toaster contains toasts.
2. Although this PR only provides support for a single toast, the code is scalable to support multiple toasts.
3. The toast is not fixed (`position: fixed`) in the Wix Studio template for desktop. To prevent messages from being missed when a page is scrolled down, I decided to make the toast fixed for all devices.
4. The board has been created. It contains different types of toasts to give users the ability to redesign them. They are not Variants, therefore not added to the UI Kit group.

**What is not done** in this PR:
1. Support for multiple toasts.
2. Custom toast duration.

### Board

Screenshot 2024-10-17 at 14 58 06

### Using toast API

```
const { runToast } = useToaster();
...
runToast({ type: ToastType.Success, message: 'The item is added to the cart' });
```

Enjoy the video:

https://github.com/user-attachments/assets/71aa3670-61f9-4c28-bac7-aaeedf67a5d9

### Switching between page and cart

I keep the message. 
(I press ESC on video)

https://github.com/user-attachments/assets/54e91370-582e-4a22-be65-3de0ebdc6a72

### Responsive

| Mobile | Tablet |
|----------|----------|
| Screenshot 2024-10-17 at 19 07 08 | Screenshot 2024-10-17 at 19 06 53 |

Desktop you've seen in the videos.

### DIFF
--- a/src/components/toaster/toast/toast.module.scss
+++ b/src/components/toaster/toast/toast.module.scss
@@ -14,7 +14,7 @@
     gap: 14px;
 }
 
-.root.cart {
+.root.inCart {
     inset: auto 0 0 auto;
     width: 100%;
     max-width: var(--drawerMaxWidth);
--- a/src/components/toaster/toast/toast.module.scss
+++ b/src/components/toaster/toast/toast.module.scss
@@ -1,0 +1,58 @@
+@import '@styles/variables.scss';
+
+.root {
+    position: absolute;
+    top: var(--headerHeight);
+    padding: 14px;
+    width: 480px;
+    color: var(--white);
+    pointer-events: all;
+
+    display: flex;
+    justify-content: space-between;
+    align-items: start;
+    gap: 14px;
+}
+
+.root.cart {
+    inset: auto 0 0 auto;
+    width: 100%;
+    max-width: var(--drawerMaxWidth);
+}
+
+@media (max-width: $desktop-width) {
+    .root {
+        top: auto;
+        bottom: 0;
+        width: 100%;
+    }
+}
+
+.root.error {
+    background-color: var(--error);
+}
+
+.root.warning {
+    background-color: var(--colorD2);
+}
+
+.root.success {
+    background-color: var(--colorE4);
+}
+
+.root.info {
+    background-color: var(--colorC3);
+}
+
+.description {
+    font: var(--paragraph3);
+    flex-grow: 1;
+}
+
+.close {
+    flex-shrink: 0;
+}
+
+.closeIcon {
+    cursor: pointer;
+}
--- a/src/components/toaster/toast/toast.tsx
+++ b/src/components/toaster/toast/toast.tsx
@@ -1,0 +1,45 @@
+import * as RadixToast from '@radix-ui/react-toast';
+import classNames from 'classnames';
+import { CloseIcon } from '../../icons';
+
+import styles from './toast.module.scss';
+import { ToastType } from '../toaster-context';
+
+interface ToastProps {
+    children: React.ReactNode;
+    type: ToastType;
+    isOpen: boolean;
+    location?: string;
+    className?: string;
+    onOpenChange: (isOpen: boolean) => void;
+    onClose: () => void;
+}
+
+export const Toast = (props: ToastProps) => {
+    const { children, type, isOpen, location, className, onOpenChange, onClose } = props;
+
+    return (
+        <RadixToast.Root
+            className={classNames(
+                styles.root,
+                styles[type],
+                location && styles[location],
+                className,
+            )}
+            open={isOpen}
+            onOpenChange={onOpenChange}
+        >
+            <RadixToast.Description className={styles.description}>
+                {children}
+            </RadixToast.Description>
+            <RadixToast.Close asChild className={styles.close}>
+                <CloseIcon
+                    width={24}
+                    className={styles.closeIcon}
+                    onClick={onClose}
+                    aria-label="Close"
+                />
+            </RadixToast.Close>
+        </RadixToast.Root>
+    );
+};
--- a/src/components/toaster/toast/toast.tsx
+++ b/src/components/toaster/toast/toast.tsx
@@ -3,11 +3,10 @@ import classNames from 'classnames';
 import { CloseIcon } from '../../icons';
 
 import styles from './toast.module.scss';
-import { ToastType } from '../toaster-context';
 
 interface ToastProps {
     children: React.ReactNode;
-    type: ToastType;
+    type: string;
     isOpen: boolean;
     isCartOpen?: boolean;
     className?: string;
--- a/src/components/toaster/toast/toast.tsx
+++ b/src/components/toaster/toast/toast.tsx
@@ -9,7 +9,7 @@ interface ToastProps {
     children: React.ReactNode;
     type: ToastType;
     isOpen: boolean;
-    isCartOpen?: string;
+    isCartOpen?: boolean;
     className?: string;
     onOpenChange: (isOpen: boolean) => void;
     onClose: () => void;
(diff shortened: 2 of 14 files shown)


src/hooks/use-cart.ts
```diff
@@ -22,28 +25,54 @@ export const useCart = () => {
...
-        triggerUpdateItemQuantity({ id, quantity }).finally(() => {
-            setUpdatingCartItems((prev) => prev.filter((itemId) => itemId !== id));
-        });
+        triggerUpdateItemQuantity({ id, quantity })
+            .catch((error) => {
+                runToast({
+                    type: ToastType.Error,
+                    message: `Failed to update the item quantity. ${getErrorMessage(error)}`,
+                });
+                throw error;
```
alisey, line 34: `throw error;` - why?

src/components/toaster/toast/toast.module.scss
```diff
@@ -0,0 +1,58 @@
...
+    display: flex;
+    justify-content: space-between;
+    align-items: start;
+    gap: 14px;
+}
+
+.root.inCart {
+    inset: auto 0 0 auto;
+    width: 100%;
+    max-width: var(--drawerMaxWidth);
```
alisey, line 20: ?! Implicit dependency on a variable defined in a random component's file.

Remove this. Why does the toast component know about cart.

src/components/toaster/toast/toast.module.scss
```diff
@@ -0,0 +1,58 @@
...
+}
+
+@media (max-width: $desktop-width) {
+    .root {
+        top: auto;
+        bottom: 0;
+        width: 100%;
```
alisey, line 27: So it will be wider on tablets than on desktop?

src/components/toaster/toaster.module.scss
```diff
@@ -0,0 +1,10 @@
+@import '@styles/variables';
+
+.viewport {
+    position: fixed;
+    inset: 0;
+    display: flex;
+    justify-content: center;
+    pointer-events: none;
+    z-index: 5;
```
alisey, line 9: What's 5?

src/components/toaster/toaster-context.tsx
```diff
@@ -0,0 +1,61 @@
...
+    const onOpenChange = () => (isOpen: boolean) => {
+        setToast((prevToast) => (prevToast ? { ...prevToast, isOpen } : null));
+    };
+
+    const providerValue = { toast, runToast, closeToast, onOpenChange };
+    return <ToasterContext.Provider value={providerValue}>{children}</ToasterContext.Provider>;
+}
+
+interface ToasterContextType {
+    toast: ToastData | null;
```
alisey, line 47: What's this?

src/components/toaster/toaster-context.tsx
```diff
@@ -0,0 +1,61 @@
...
+    runToast: (toastData: { type: ToastType; message: string; duration?: number }) => void;
```
alisey, line 48: Just `toast`

src/components/toaster/toaster-context.tsx
```diff
@@ -0,0 +1,61 @@
...
+    closeToast: () => void;
```
alisey, line 49: > Although this PR only provides support for a single toast, the code is scalable to support multiple toasts.

It's not.

src/components/toaster/toaster-context.tsx
```diff
@@ -0,0 +1,61 @@
...
+    onOpenChange: (isOpen: boolean) => void;
```
alisey, line 50: > Although this PR only provides support for a single toast, the code is scalable to support multiple toasts.

With methods like this, it's not scalable.

src/components/toaster/toast/toast.tsx
```diff
@@ -0,0 +1,37 @@
+import * as RadixToast from '@radix-ui/react-toast';
+import classNames from 'classnames';
+import { CloseIcon } from '../../icons';
+
+import styles from './toast.module.scss';
+import { ToastType } from '../toaster-context';
```
alisey, line 6: Put together with other typescript imorts.

src/components/toaster/toast/toast.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+interface ToastProps {
+    children: React.ReactNode;
+    type: ToastType;
+    isOpen: boolean;
+    isCartOpen?: boolean;
```
alisey, line 12: ?!

src/components/toaster/toaster-context.tsx
```diff
@@ -0,0 +1,61 @@
...
+    Warning = 'warning',
+    Info = 'info',
+}
+
+interface ToastData {
+    type: ToastType;
+    message: string;
+    isOpen: boolean;
+    timeout: ReturnType<typeof setTimeout>;
+    origin?: string;
```
alisey, line 17: What's `origin`? Why does the toast need to know its origin?

src/components/toaster/toaster-context.tsx
```diff
@@ -0,0 +1,61 @@
...
+            if (toast && toast.timeout) clearTimeout(toast.timeout);
+            return null;
+        });
+    };
+
+    const onOpenChange = () => (isOpen: boolean) => {
+        setToast((prevToast) => (prevToast ? { ...prevToast, isOpen } : null));
+    };
+
+    const providerValue = { toast, runToast, closeToast, onOpenChange };
```
alisey, line 42: It's not provider value, it's context value.

src/hooks/use-cart.ts
```diff
@@ -22,28 +25,54 @@ export const useCart = () => {
...
 
     const checkout = async () => {
         const checkoutResponse = await ecomAPI.checkout();
 
         if (checkoutResponse.status === 'success') {
             window.location.href = checkoutResponse.body.checkoutUrl;
         } else {
-            alert('checkout is not configured');
+            runToast({
+                type: ToastType.Error,
```
alisey, line 72: 1. Just pass strings for type, does TypeScript allow using strings here?

2. I'm not sure it's a good idea to put toasts into the useCart hook, rather than calling `checkout().catch(() => toast.error('Oops'))`. If the template author wants to show the error in the form. @Letushev what do you think?

src/hooks/use-cart.ts
```diff
@@ -22,28 +25,54 @@ export const useCart = () => {
...
+                message: `Checkout is not configured. ${getErrorMessage(checkoutResponse.error)}`,
```
alisey, line 73: > Checkout is not configured.

You don't know that. The user could be offline.

alisey (commented):
Move toast to a separate PR. Forget that cart exists, implement the toast component and related hooks. Then use it in the cart in a separate PR.

src/hooks/use-cart.ts
Letushev: I think it shouldn't be part of `useCart` hook.
>  If the template author wants to show the error in the form

True. Also other templates might want to use different toast message or duration.

src/hooks/use-cart.ts
tetianaman: _use-cart_
`const addToCart = (args) => triggerAddToCart(args).catch((error) => runToast(...));`

usage:
`addToCart(args).then(() => cartOpener.setIsOpen(true));`

In this case `then` executes. I should have thrown error in `catch`.

But never mind, I'll move toast calls out of `useCart` hook.

src/components/toaster/toast/toast.module.scss
tetianaman: Yes, I got this waking up at midnight, that cart can be quite different in projects and I should tell toast its location. Not only location. For e.g. `width` as well.

src/components/toaster/toast/toast.module.scss
tetianaman: yes. The same as in Wix Studio template

src/components/toaster/toaster.module.scss
tetianaman: Toast should over the cart. Setting '5' I reserve a few values in case there will be some objects over the cart but under the toast. What is the best practice to choose `z-index`?

PR #157, opened 2024-10-17
refactor: move templates shared stuff to lib folder

Letushev:


_codux/boards/color-select.board.tsx
```diff
@@ -1,6 +1,6 @@
 import { useState } from 'react';
 import { createBoard } from '@wixc3/react-board';
-import { ColorSelect, ColorSelectOption } from '~/components/color-select/color-select';
+import { ColorSelect, ColorSelectOption } from 'lib/components/color-select/color-select';
```
yurii-ve, line 3: not sure about moving components to the lib.. might be confusing for the user why some components are in `lib` and some in `src`. For example `Accordion` sounds like generic component and it's in the `src`, on the other hand - `RangeSlider`, also looks like a generic component but it's in the `lib`.

Letushev:
@alisey @yurii-ve @olehrakwix PR is ready for review
- Renamed `lib/api` to `lib/ecom`
- Moved `useCart` to `lib/ecom/hooks`
- Changed functions to get data in route loaders
- Added index files to `lib` subfolders except for `lib/components`

olehrakwix (approved):

_codux/board-wrappers/component-wrapper.tsx
```diff
@@ -1,6 +1,6 @@
 import { createRemixStub } from '@remix-run/testing';
 import { PropsWithChildren } from 'react';
-import { EcomAPIContextProvider } from '~/api/ecom-api-context-provider';
+import { EcomAPIContextProvider } from 'lib/ecom';
```
alisey, line 3: Oh, I didn't realize imports are going to look like this, `lib` looks like a name of an NPM package. For some reason I though it's going to be `~/lib/ecom`.

Maybe we should do this?
`~/src/components`
`~/lib/components`

alisey: Even without `lib`, `~/src/router` is more intuitive for me than `~/router`

alisey (approved):

_codux/board-wrappers/component-wrapper.tsx
Letushev: Updated all imports to the pattern
`~/src/components`
`~/lib/components`

PR #160, opened 2024-10-22
feat: load more products

olehrakwix:
- add ability to load more products if not all shown on page
- `Load More` button is shown only if there is more products available
- the url stays unchanged, so we don't need to handle additional tricky cases when user loads a list from non-first page
- once filters or sort order changed, loaded projects reset and we are showing only first chunk of products
- added hook `useProductSorting` to easily access current products sort order
- added hook `useProductSlice` that is handling paged products loading

in order to demonstrate how it works, I changed a default page size to 3 items. in this PR default page size remains unchanged as 100 items

https://github.com/user-attachments/assets/72868edb-e1ed-418c-97cb-4ec9d28b4559


lib/hooks/use-products-list.ts
```diff
@@ -0,0 +1,54 @@
...
+                filters: appliedFilters,
+                sortBy: productsSortBy,
+                skip: products.length,
+            });
+
+            if (nextProductsResponse.status === 'success') {
+                setProducts((prev) => [...prev, ...nextProductsResponse.body.items]);
+                setTotalProductsCount(nextProductsResponse.body.totalCount);
+            }
+        } finally {
```
yurii-ve, line 42: maybe add `catch` with just `console.log` or `alert` for now.. and once toasts functionality is merged we'll show the toast.

olehrakwix: done

lib/hooks/use-products-list.ts
```diff
@@ -0,0 +1,58 @@
+import { SerializeFrom } from '@remix-run/node';
+import { useEffect, useState } from 'react';
+import { getEcomApi, Product } from '../ecom';
+import { useAppliedProductFilters } from './use-applied-product-filters';
+import { useProductSorting } from './use-product-sorting';
+
+export function useProductsList(
```
alisey, line 7: This hook is confusing. It takes the category slug, so I would expect it to return all products from that category. Oh, but it also takes the `initialTotalCount`. What could that possibly mean? I can understand the initial count, I can understand the total count. I cannot understand the initial total count.

Only after looking at the implementation of the hook can I figure out that the hook doesn't actually return all products from the category, it also takes sorting and filtering params into account. But unlike the category slug, it takes them from the URL.

There's also suspicious:
```
    const [products, setProducts] = useState(initialProducts);
    useEffect(() => {
        setProducts(initialProducts);
    }, [initialProducts]);
```

I can't figure out if it's synchronized with the rest of the state in the hook, like when I've loaded a few pages and initialProducts have changed. Why does the hook care if `initialProducts` have changed.

---

It's not "products list", it "product list". It's always singular when you have a list of something.

lib/hooks/use-products-list.ts
```diff
@@ -0,0 +1,58 @@
...
+                filters: appliedFilters,
+                sortBy: productsSortBy,
+                skip: products.length,
+            });
+
+            if (nextProductsResponse.status === 'success') {
+                setProducts((prev) => [...prev, ...nextProductsResponse.body.items]);
+                setTotalProductsCount(nextProductsResponse.body.totalCount);
+            } else {
+                console.error(nextProductsResponse.error);
```
yurii-ve, line 42: `throw` here

lib/hooks/use-products-list.ts
yurii-ve: of course it's temporary, but maybe alert is better.. at least user will see something

app/routes/products.$categorySlug/route.tsx
```diff
@@ -39,12 +40,15 @@ export default function ProductsPage() {
     const { category, categoryProducts, allCategories, productPriceBounds } =
         useLoaderData<typeof loader>();
 
+    const { products, totalProductsCount, isLoadingProducts, canLoadMore, loadMore } =
+        useProductsList(category.slug!, categoryProducts.items, categoryProducts.totalCount);
```
alisey, line 44: There's `categoryProducts` and there's `products`, the difference is unclear.

lib/hooks/use-products-list.ts
olehrakwix: - updated hook to accept all parameters form outside (filters, sort order, category slug)
- renamed to `useProductSlice`, I think this has more sense in our case
- added jsDoc comment where trying to explane better the purpose of this hook

in general, this is done in such a way, cause we have initial data from loader on first page render, but then if user changes filters or sort order, remix loads a new data from client side (without full page reload) and updates the result of `useLoaderData` hook.  this is why we need this strange `useEffect`

alisey:
I have two proposals for the API.

1. Basically your `slice` idea but with different naming.

```ts
const {
    products,
    totalProducts,
    loadMoreProducts,
    isLoadingMoreProducts,
} = useProductResults({ categorySlug, filters, sorting, initialResultsFromLoader });
```

2. Bake everything into one uber-hook.
```ts
const {
    allCategories,
    category,
    priceBounds,
    currency,
    products,
    totalProducts,
    loadMoreProducts,
    isLoadingMoreProducts,
} = useProductsPageData(useLoaderData<typeof loader>());
```

We could even move `useLoaderData` into the hook, and it would become `const { ... } = useProductsPageData<typeof loader>()`, but I think it's generally nicer when `useLoaderData` is only used in the route component.

olehrakwix:
@alisey, I do like your first proposal more, because I also believe that `useLoaderData` shouldn't be used anywhere outside the route component.
will update PR

lib/ecom/api.ts
```diff
@@ -78,7 +78,7 @@ function createApi(): EcomAPI {
     const wixClient = getWixClient();
 
     return {
-        async getProductsByCategory(categorySlug, { limit, filters, sortBy } = {}) {
+        async getProductsByCategory(categorySlug, { skip, limit, filters, sortBy } = {}) {
```
alisey, line 81: Let's specify default values here.

lib/hooks/use-product-results.ts
```diff
@@ -0,0 +1,62 @@
...
+export interface UseProductResultsArgs {
+    categorySlug: string;
+    filters: IProductFilters;
+    sorting: ProductSortBy;
+    initialResultsFromLoader: Slice<Product | SerializeFrom<Product>>;
+}
+
+/**
+ * Manages product slicing and loads next slices from the client side
+ * without a full page reload. Syncs with initialResultsFromLoader.
```
alisey, line 15: useProductsPageResults()

Returns the list of products displayed on the products page.


initialResultsFromLoader - the first batch of results loaded by the route loader to avoid fetching it again client-side.

lib/hooks/use-product-results.ts
```diff
@@ -0,0 +1,62 @@
...
+ */
+export function useProductResults({
+    categorySlug,
+    filters,
+    sorting,
+    initialResultsFromLoader,
+}: UseProductResultsArgs) {
+    const [productSlice, setProductSlice] = useState(initialResultsFromLoader);
+    useEffect(() => {
```
alisey, line 24: `initialResultsFromLoader` will change when the filters or the category has been changed, and the route returned the first batch of results, and we need to discard the subsequent batches that we loaded client-side.

lib/hooks/use-product-results.ts
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Manages product slicing and loads next slices from the client side
+ * without a full page reload. Syncs with initialResultsFromLoader.
+ */
+export function useProductResults({
+    categorySlug,
+    filters,
+    sorting,
+    initialResultsFromLoader,
```
alisey, line 21: resultsFromLoader

alisey (approved):

lib/hooks/use-product-results.ts
alisey: Maybe there are better options, but here's one:

1. Have a `const counter = useRef(0);`
2. In `dataFromLoader` effect: `counter++`
3. In `onUnmount` effect: `counter++`
4. In `loadMore`: read the counter before making the request, after the request has completed, compare the current value of the counter to the previous. If they don't match, ignore the result of the request.

olehrakwix: fixed, also I think name `resultsFromLoaderVersionRef` makes more sense for this counter

app/routes/products.$categorySlug/route.tsx
```diff
@@ -40,20 +41,28 @@ export const handle = {
...
-        sorting,
-        initialResultsFromLoader,
-    });
+    const resultsFromLoader = useMemo<ProductsPageResults>(
+        () => ({
+            products: productsFromLoader.items,
+            totalProducts: productsFromLoader.totalCount,
+        }),
+        [productsFromLoader],
+    );
```
alisey, line 58: Why not just pass it directly without renaming properties?

alisey (approved):

app/routes/products.$categorySlug/route.tsx
olehrakwix: I thought `products` is a little bit better name then `items`, but both are fine for me, fixed

PR #162, opened 2024-10-25
feat: user session

olehrakwix:
- implement user session using remix-way
- store wix headless API access tokens in session
- store wix client id in session
- root loader is responsible to generate access tokens and set a session cookie if needed
- each time some loader needs API it is iniaialized from current request (using wix client token from session cookie)
- session is reset once wix client id changed
- now both front-end and back-end uses same access tokens and acts in context of same wix user (anonymous visitor for now)


VladyslavG:
>session and API is initialized by root loader, so this happens on each request
not really, root loader works only when we reload page or when we change URL on a root level. Am I wrong ?

.env.template
```diff
@@ -1 +1,2 @@
 WIX_CLIENT_ID=35a15d20-4732-4bb8-a8ef-194fd1166827
+SESSION_COOKIE_SIGNING_SECRET=SOME_SECRET
```
VladyslavG, line 2: is this secret set by the wix-headless feature when we connect to wx?

lib/ecom/index.ts
```diff
@@ -1,5 +1,4 @@
 export * from './api-context';
-export * from './api';
```
VladyslavG, line 2: why this ?

lib/ecom/session.ts
```diff
@@ -0,0 +1,42 @@
...
+
+type SessionData = {
+    wixEcomTokens: Tokens;
+    wixClientId: string;
+};
+
+const { getSession, commitSession } = createCookieSessionStorage<SessionData, void>({
+    cookie: {
+        name: '__session',
+        maxAge: 3600 * 24 * 30 * 3, // 3 months
```
VladyslavG, line 13: only if containing February :) in non-leap year

lib/ecom/session.ts
```diff
@@ -0,0 +1,42 @@
...
+        secure: true,
+        httpOnly: true,
+        sameSite: 'lax',
+
+        secrets: [process.env.SESSION_COOKIE_SIGNING_SECRET || 's3cret1'],
```
VladyslavG, line 18: is this newline on purpose ?

VladyslavG: || -> ??

app/root.tsx
```diff
@@ -30,12 +32,24 @@ export const meta: MetaFunction = () => {
     return [{ title: 'ReClaim: Home Goods Store' }];
 };
 
-export async function loader() {
-    return {
-        ENV: {
-            WIX_CLIENT_ID: process?.env?.WIX_CLIENT_ID,
+export async function loader({ request }: LoaderFunctionArgs) {
+    const { wixEcomTokens, setCookie } = await initializeSession(request);
```
yurii-ve, line 36: Relationship between `initializeSession` and `initializeEcomApi` look a bit tangled to me.
`initializeEcomApi` expects tokens returned by `initializeSession`. At the same time `initializeSession` internally calls `initializeEcomApi` to generate visitor tokens.

olehrakwix:
> not really, root loader works only when we reload page or when we change URL on a root level. Am I wrong ?
@VladyslavG , you are completely right. will fix this

.env.template
olehrakwix: it has completely nothing to do with wix headless

lib/ecom/session.ts
olehrakwix: > only if containing February :) in non-leap year
will update to 90 days

> why 3 months?
just my imagination. do you have some more meaningful period?

lib/ecom/session.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    const sessionWixClientId = session.get('wixClientId');
+    const wixClientId = getWixClientId();
+
+    // reset token if wix client id has changed
+    let tokens = sessionWixClientId !== wixClientId ? undefined : session.get('wixEcomTokens');
+
+    const { client } = initializeEcomApi(tokens);
+
+    let setCookie: string | undefined;
```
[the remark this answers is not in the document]
olehrakwix: this is from remix documentation. maybe setCookieHeader works for you?

    /**
     * Stores all data in the Session and returns the Set-Cookie header to be
     * used in the HTTP response.
     */

lib/ecom/session.ts
VladyslavG: 100 is a round number

lib/ecom/session.ts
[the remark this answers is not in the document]
olehrakwix: but ok, I'm fine wih `sessionCookie`

app/root.tsx
olehrakwix: it works exactly as you described. will try to somehow make it less coupled

lib/ecom/session.ts
yurii-ve: maybe we can use the lifespan of wix tokens?

lib/ecom/api.ts
```diff
@@ -37,28 +32,7 @@ function getWixClientId() {
...
-
-function getTokensClient() {
-    ensureSessionIntegrity();
-
-    const tokens = Cookies.get(WIX_SESSION_TOKEN_COOKIE_KEY);
-    return tokens ? JSON.parse(tokens) : undefined;
-}
-
-function getWixClient() {
+function createApiWixClient(tokens?: Tokens) {
```
yurii-ve, line 35: `createWixSDKClient` or `createWixClient` to avoid having `createApiWixClient` and `createApi`?

lib/ecom/session.ts
olehrakwix: changed to 100 days.
@yurii-ve ,  we can not use wix tokens at the momen of session initialization, cause this is kinda 'static' session cookie initialization, we have no wix tokens yet

lib/ecom/api-context.tsx
```diff
@@ -13,7 +14,14 @@ export const useEcomAPI = (): EcomAPI => {
...
 };
 
-export const EcomAPIContextProvider: FC<React.PropsWithChildren> = ({ children }) => {
+export interface EcomAPIContextProviderProps extends React.PropsWithChildren {
+    tokens?: Tokens;
+}
+
+export const EcomAPIContextProvider: FC<EcomAPIContextProviderProps> = ({ tokens, children }) => {
+    const client = createWixClient(tokens);
+    const api = createApi(client);
```
alisey, line 23: Shouldn't this be memoized?

lib/ecom/session.ts
```diff
@@ -0,0 +1,51 @@
...
+        httpOnly: true,
+        sameSite: 'lax',
+
+        secrets: [process.env.SESSION_COOKIE_SIGNING_SECRET ?? 's3cret1'],
+    },
+});
+
+export { commitSession };
+
+export async function initializeSession(request: Request) {
```
alisey, line 24: `initializeEcomSession`, `initializeEcomApi`?

lib/ecom/api-context.tsx
olehrakwix: nice catch, fixed this

app/root.tsx
```diff
@@ -64,25 +79,25 @@ export function Layout({ children }: React.PropsWithChildren) {
...
-        window.ENV = data.ENV;
+        window.ENV = ENV;
     }
 
     return (
-        <ContentWrapper>
-            <Outlet />
-        </ContentWrapper>
+        <EcomAPIContextProvider tokens={wixEcomTokens}>
+            <ContentWrapper>
```
yurii-ve, line 97: can't we pass `tokens` inside the `ContentWrapper`?
I don't like the idea that now `CartOpenContextProvider` is inside the `ContentWrapper`, but `EcomAPIContextProvider` is here... if there was a new context provider I wouldn't know where and what to wrap with it :)

lib/ecom/api.ts
```diff
@@ -103,6 +86,29 @@ function createApi(): EcomAPI {
                 return failureResponse(EcomApiErrorCodes.GetProductsFailure, getErrorMessage(e));
             }
         },
+        async getFeaturedProducts(categorySlug, count) {
```
yurii-ve, line 89: I don't think it should be a part of the API

alisey: Why not?

lib/ecom/session.ts
alisey: There's remix session as well.

olehrakwix: basically this function initializes ecom session always and remix session if it is not yet initialized

app/root.tsx
olehrakwix: fixed

lib/ecom/api.ts
olehrakwix: I don't see any crusial reasons why this should or shouldn't be a part of an API. so, @alisey asked to move it to API, and @yurii-ve said this shouldn't be there. lets vote? 😀

yurii-ve: My reasoning is that (if I'm not missing anything) API doesn't have notion of 'Featured Products', it's purely ui thing, and I don't like the idea of mixing API specific functionality with client code that just happen to call API.
But given that we already have `getPromotedProducts` below, maybe it's ok

alisey: `getPromotedProducts` needs to be removed. It's unused, and conceptually it's the same thing as featured products.

`EcomAPI` is our own interface. It wraps `WixClient` but doesn't map one to one, even for methods like `getOrder`.

`WixClient` lacks useful methods like `getProductPriceBounds`, so we've had to add our own. Same goes for `getFeaturedProducts`, it's basically sort by recommended, with a fallback to all products.

yurii-ve (approved):

PR #164, opened 2024-10-27
feat: support getStaticRoutes

PeterShershov:
Add `getStaticRoutes` to relevant routes in order to suggest data coming from an headless service on Codux's Pages Panel


PeterShershov:
@alisey please review, should be merged to main as `getStaticRoutes` is already supported in Codux

alisey:
> should be merged to main as getStaticRoutes is already supported in Codux

I've changed the branch to `development` to avoid merging in two directions from main to development and vice versa.

lib/ecom/api.ts
```diff
@@ -290,6 +290,15 @@ function createApi(): EcomAPI {
                 return failureResponse(EcomApiErrorCodes.GetProductsFailure, getErrorMessage(e));
             }
         },
+        async getAllProducts(limit = 100) {
```
olehrakwix, line 293: this is not going to return all projects...
in fact, SDK/API could not return more then 100 items. if there is more then 100 products, there will be an issue. to handle this correctly, we have to load items in few requests, see example here:
https://github.com/codux-templates/e-commerce-remix/commit/fb0076f6093ed1cfeebbe0d96b3596ae107830d1

app/routes/product-details.$productSlug/route.tsx
```diff
@@ -22,6 +22,16 @@ export const loader = ({ params, request }: LoaderFunctionArgs) => {
     return getProductDetailsRouteData(params.productSlug, request.url);
 };
 
+export const getStaticRoutes = async () => {
```
alisey, line 25: For a user it's going to be difficult to understand what this function is, since it's not a Remix feature. Do we have a type that we could use here to make it more clear what it is for and to help find documentation?

Something like: `export const getStaticRoutes: CoduxStaticRoutes = async`, where hovering on `CoduxStaticRoutes` provides a link to the documentation. Even without a link knowing that it's related to Codux would help already.

app/routes/product-details.$productSlug/route.tsx
```diff
@@ -22,6 +22,16 @@ export const loader = ({ params, request }: LoaderFunctionArgs) => {
...
+    const products = await getEcomApi().getAllProducts();
+
+    if (products.status === 'failure') {
+        throw new Error(`${products.error.code} ${products.error.message || ''}`);
```
alisey, line 29: I'll make sure to refactor this so message isn't optional.

If the message isn't optional, do you still need the code?
* If yes, let's have something like `formatEcomApiError(products.error)`
* If not, let's keep it as is, and I'll change to `new Error(products.error.message)` or ideally `throw prodcuts.error`;

app/routes/product-details.$productSlug/route.tsx
```diff
@@ -22,6 +22,16 @@ export const loader = ({ params, request }: LoaderFunctionArgs) => {
...
+    }
+
+    return products.body.map((product) => `/product-details/${product.slug}`);
```
alisey, line 32: `ROUTES.productDetails`? If we're keeping it.

lib/ecom/api.ts
alisey: @olehrakwix it does not need to return all products, it needs to return some products.

We could use `await api.getProductsByCategory('all-products');` instead. I'm not a huge fan of hardcoding the string `"all-products"` but it's hardcoded in many places.

olehrakwix: ok, I was under impression that we are going to collect all available on site routes with this function, but if we need just "some" products - then you are correct

app/routes/product-details.$productSlug/route.tsx
alisey: The `error.message` property is always defined now, see #165.

Concatenation of error code and message will results in things like "ProductNotFound Product not found".

This should be fine now: `throw new Error(products.error.message)`

app/routes/product-details.$productSlug/route.tsx
alisey: Thinking about where this type would be imported from... `@wixc3/define-remix-app`?

It's a dev dependency, but since we'd be only importing a type, it should be fine.

PeterShershov: Nice one! On it.

lib/ecom/api.ts
PeterShershov: I'll consult with Omer and Nadav about this behavior and update here.

package.json
```diff
@@ -42,10 +42,10 @@
...
     "@types/js-cookie": "^3.0.6",
     "@types/lodash.debounce": "^4.0.9",
-    "@types/react": "^18.3.12",
-    "@types/react-dom": "^18.3.1",
-    "@typescript-eslint/eslint-plugin": "^8.12.1",
-    "@wixc3/define-remix-app": "^4.5.0",
+    "@types/react": "^18.3.11",
+    "@types/react-dom": "^18.3.0",
+    "@typescript-eslint/eslint-plugin": "^8.8.0",
+    "@wixc3/define-remix-app": "^4.5.1",
```
alisey, line 48: Only `"@wixc3/define-remix-app"` should be bumped, the rest should be left alone.

PeterShershov: oops, bad merge.

app/routes/product-details.$productSlug/route.tsx
```diff
@@ -14,13 +14,24 @@ import { ProductPrice } from '~/src/components/product-price/product-price';
...
+import type { GetStaticRoutes } from '@wixc3/define-remix-app';
 
 import styles from './route.module.scss';
 
 export const loader = ({ params, request }: LoaderFunctionArgs) => {
     return getProductDetailsRouteData(params.productSlug, request.url);
 };
 
+export const getStaticRoutes: GetStaticRoutes = async () => {
+    const products = await getEcomApi().getAllProducts();
```
alisey, line 26: @olehrakwix now when `getEcomApi()` is gone, what should be used here?

@PeterShershov in which environment does `getStaticRoutes` run? Preview?

olehrakwix: in this particular case, please use:
```
const api = createApi(createWixClient());
```

alisey: Updated.

@olehrakwix I don't like that from the user's POV there are two completely different ways to get the API instance: `createApi(createWixClient())` and `initializeEcomApi()`.

Let's discuss this.

alisey (approved):

lib/ecom/api.ts
alisey: I checked `api.getProductsByCategory`, it makes two requests: category first, and then the list of products in it. I kept `getAllProducts`, but renamed it to `getProducts`. I think I'll combine these two functions later, and make category an optional parameter.

app/routes/product-details.$productSlug/route.tsx
PeterShershov: @alisey `getStaticRoutes` is being called on the server.
see https://github.com/wixplosives/codux-core/blob/98d7de6e9bb6e4975068cd6f3ea1860082903ea9/packages/define-remix-app/src/define-remix-app.tsx#L280

PR #167, opened 2024-10-29
refactor: Remove route path constants

yurii-ve:
Per Nadav's request removed `/src/router/config.ts` and hardcoded all the paths as string in the places where they are needed.


olehrakwix (approved):
approving with a pain in my heart 🥲

alisey (approved):

yurii-ve:
> approving with a pain in my heart 🥲

I feel you

PR #168, opened 2024-10-29
security(deps): upgrade to latest stable

AviVahl:
- reduces audit failures. still bunch coming from `@wix/sdk` and `remix`
- regenerate lock file


alisey:
Codux does not support `@use`. @ynesteruk am I right?

alisey (changes requested):
Revert changes to `@import`, since Codux doesn't support `@use`.

ynesteruk:
> Codux does not support @use. @ynesteruk am I right?

Yes. We don't treat it correctly, especially when it comes to collecting variables/class names from scss files

alisey (approved):

PR #174, opened 2024-10-30
Add members area pages

yurii-ve:
Added `members-area` page and two 'under construction' sub-pages: `members-area/my-account` and `members-area/my-orders`. These pages aren't accessible through the UI for now.

image

### DIFF
--- a/app/routes/members-area.my-account/route.module.scss
+++ b/app/routes/members-area.my-account/route.module.scss
@@ -1,0 +1,4 @@
+.page {
+    text-align: center;
+    font: var(--heading5);
+}
--- a/app/routes/members-area.my-account/route.module.scss
+++ b/app/routes/members-area.my-account/route.module.scss
@@ -1,4 +1,4 @@
-.page {
+.underConstructionMessage {
     text-align: center;
     font: var(--heading5);
 }
--- a/app/routes/members-area/route.module.scss
+++ b/app/routes/members-area/route.module.scss
@@ -1,0 +1,27 @@
+.page {
+    padding-top: 100px;
+    width: 980px;
+    margin: 0 auto;
+}
+
+.menu {
+    display: flex;
+    border-bottom: 1px solid var(--secondary1);
+}
+
+.menuItem {
+    padding: 14px 30px;
+    border-bottom: 4px solid transparent;
+}
+
+.menuItem:hover {
+    border-bottom-color: var(--secondary2);
+}
+
+.menuItem.active {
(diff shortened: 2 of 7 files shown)


olehrakwix (approved):

app/routes/members-area.my-account/route.module.scss
```diff
@@ -0,0 +1,4 @@
+.page {
+    text-align: center;
+    font: var(--heading5);
```
alisey, line 3: Are you sure about `--heading5` as the body text?

app/routes/members-area/route.tsx
```diff
@@ -0,0 +1,26 @@
...
+    return (
+        <div className={styles.page}>
+            <div className={styles.menu}>
+                <NavLink to={'/members-area/my-account'} className={menuItemClassName}>
+                    My Account
+                </NavLink>
+                <NavLink to={'/members-area/my-orders'} className={menuItemClassName}>
+                    My Orders
+                </NavLink>
+            </div>
```
alisey, line 20: These are tabs. Not a menu. It'd make sense to create a Tabs component or at least reusable HTML classes for building a tab bar.

app/routes/members-area.my-account/route.module.scss
yurii-ve: it's temporary for the "This page is under construction" message and going to removed when the actual page is implemented. But I can wrap the message in another div and apply the style there, for this to make more sense.

alisey (approved):

src/styles/common.scss
```diff
@@ -360,6 +360,43 @@
     }
 }
 
+/* Tabs **************************************************************/
+
+.tabs {
+    display: flex;
+    border-bottom: 1px solid var(--secondary1);
+    overflow: scroll;
```
alisey, line 368: Should be `auto`, I guess. On Windows `overflow: scroll` means there's always a scrollbar, AFAIK.

app/routes/members-area/route.module.scss
```diff
@@ -0,0 +1,10 @@
+.page {
+    padding-top: 100px;
+    width: 100%;
```
alisey, line 3: Is `width: 100%` necessary? If this has to be added to every page, then it would make more sense to remove the container style that prevents automatic stretching.

alisey: Ah. Is it because of `margin: 0 auto`?

PR #175, opened 2024-10-30
refactor: functions for ecom api initialization

alisey:
We use several ways to get an instance of EcomAPI:

1. `initializeEcomApi(request)`
2. `createApi(createWixClient(tokens))`
3. `createApi(createWixClient())`

To unify them and to make it explicit when we're creating an anonymous instance that doesn't have a client token, I've changed these as follows:

1. `initializeEcomApiForRequest()`
2. `initializeEcomApiWithTokens()`
3. `initializeEcomApiAnonymous()`

Also made the function `createApi()` private.

### DIFF
--- a/lib/ecom/session.ts
+++ b/lib/ecom/session.ts
@@ -1,6 +1,6 @@
 import { createCookieSessionStorage } from '@remix-run/node';
 import { Tokens } from '@wix/sdk';
-import { createApi, createWixClient, getWixClientId } from './api';
+import { createEcomApi, createWixClient, getWixClientId } from './api';
 
 export type SessionData = {
     wixEcomTokens: Tokens;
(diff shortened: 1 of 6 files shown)


lib/ecom/session.ts
```diff
@@ -43,9 +48,8 @@ export async function initializeEcomSession(request: Request) {
     return { wixEcomTokens, session, shouldUpdateSessionCookie };
 }
 
-export async function initializeEcomApi(request: Request) {
+export async function initializeEcomApiForRequest(request: Request) {
     const { session } = await initializeEcomSession(request);
     const tokens = session.get('wixEcomTokens');
-    const client = createWixClient(tokens);
-    return createApi(client);
+    return tokens ? initializeEcomApiWithTokens(tokens) : initializeEcomApiAnonymous();
```
alisey, line 54: @olehrakwix is it an expected scenario that we won't have tokens here?

olehrakwix: we are getting tokens from root loader, but if it fails and ErrorBoundary will render a page - we will not have tokens there

olehrakwix (approved):

lib/ecom/session.ts
alisey: Should we throw then instead of initializing for anonymous user?

olehrakwix: I don't think so... in that case we will throw error from ErrorBoundary that will break the whole app I believe.

PR #178, opened 2024-10-30
feat: member login

olehrakwix:
image
image


lib/ecom/api.ts
```diff
@@ -282,6 +285,46 @@ function createEcomApi(wixClient: WixApiClient): EcomAPI {
...
+                responseMode: 'query',
+            });
+
+            return { oAuthData, authUrl };
+        },
+        async logout(returnUrl: string) {
+            const result = await wixClient.auth.logout(returnUrl);
+            return result;
+        },
+        async handleLoginCallback(url, oAuthData) {
```
yurii-ve, line 308: this method feels out of place here... maybe it's better to have to separate methods: `getMemberTokens` and `setMemberTokens`?

lib/ecom/session.ts
```diff
@@ -12,7 +12,14 @@ export type SessionData = {
     wixClientId: string;
 };
 
-const { getSession, commitSession } = createCookieSessionStorage<SessionData, void>({
+type FlashData = {
```
yurii-ve, line 15: why `Flash`?

yurii-ve: I see... the name defined in the `createCookieSessionStorage` 🤔

lib/ecom/api.ts
olehrakwix: we also using `wixClient.auth.parseFromUrl(url, 'query');` for this.
so, it would be like:

```
const api = initializeEcomApi();

const oAuthData = api.parseOauthDataFromUrl(request.url);

const originalOauthData = session.get("oAuthData");

const memberTokens = api.getMemberTokens(oAuthData, originalOAuthData);

api.setMemberTokens(memberTokens)

```

I don't think this makes a lot of sense. in that case, instead of incapsulate logic inside single place we are spreading it across application

lib/ecom/session.ts
olehrakwix: yes. you can dafine data and flash data for session. data - persisted for a long time. flash data - persisted until first read

lib/hooks/use-user-info.ts
```diff
@@ -0,0 +1,18 @@
...
+
+    const isUserLoggedIn = useMemo(() => api.isUserLoggedIn(), [api]);
+
+    useEffect(() => {
+        if (api.isUserLoggedIn()) {
+            api.getUser().then((u) => setUser(u));
+        }
+    }, [api]);
+
+    return { user, isUserLoggedIn };
```
yurii-ve, line 17: we can just return user info.. no info = not logged in

src/components/avatar/avatar.tsx
```diff
@@ -0,0 +1,20 @@
...
+export interface AvatarProps {
+    imageSrc: string | undefined;
+    altText: string;
+}
+
+export const Avatar = ({ imageSrc, altText }: AvatarProps) => {
+    return (
+        <div className={styles.root}>
+            {imageSrc ? (
+                <img alt={altText} src={imageSrc}></img>
```
yurii-ve, line 14: ``

src/components/user-login/user-login.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { Select, SelectItem } from '../select/select';
+
+import styles from './user-login.module.scss';
+
+enum LoginMenuAction {
+    initial = '',
+    logout = 'logout',
+}
+
+export const UserLogin = () => {
```
yurii-ve, line 14: UserMenu

src/components/header/header.module.scss
```diff
@@ -4,6 +4,8 @@
     --headerTopBarHeight: 50px;
     --headerNavigationHeight: 54px;
     --headerHeight: calc(var(--headerTopBarHeight) + var(--headerNavigationHeight));
+
+    --actionsWidth: 500px;
```
yurii-ve, line 8: isn't this too big on smaller screens?
We hide site menu if viewport is less than 1024.. on such viewport I don't think that are 500px to the right of the site menu.

lib/hooks/use-user-info.ts
olehrakwix: not exaclty. we can immediatelly know if user is logged in, but get user information requires a call to API.

src/components/header/header.module.scss
yurii-ve: I mean here 
image

lib/hooks/use-user-info.ts
yurii-ve: oh.. I see. I thought `isUserLoggedIn` is an API call as well, but it's not

let's rename to just `isLoggedIn` (in returned object).. it's clear that it's user

lib/ecom/api.ts
yurii-ve: but what does `handleLoginCallback` (don't even look at the implementation) have to do with the API? Maybe it uses API but it should not be part of the API.
Why not having it in place where it's used: 
```
const handleLoginCallback = () => {
   ... // the code from your comment here
}
```

> instead of incapsulate logic inside single place we are spreading it across application

in my opinion it's not spreading it across the application, it's separation of concerns.. sorry for my academic background 😄

app/routes/login-callback/route.tsx
```diff
@@ -0,0 +1,25 @@
...
+    const { memberTokens, returnUrl } = await api.handleLoginCallback(request.url, oAuthData);
+    if (memberTokens) {
+        session.set('wixEcomTokens', memberTokens);
+    }
+
+    const redirectUrl = new URL(returnUrl);
+    // add some query string to redirect url, because netlify in production mode
+    // keeps query string even if it is not present in redirect location
+    // and this results in `code` and `state` params presence in URL after login
+    redirectUrl.searchParams.set('logged-in', 'true');
```
olehrakwix, line 18: https://answers.netlify.com/t/netlify-keeps-searchparams-when-redirecting-in-my-next-js-app/108590

src/components/user-login/user-login.tsx
```diff
@@ -0,0 +1,60 @@
...
+            navigate(`/logout?returnPath=${location.pathname}`);
+        }
+    };
+
+    const avatar = (
+        <Avatar imageSrc={user?.profile?.photo?.url} altText={user?.profile?.nickname || ''} />
+    );
+
+    return (
+        <div className={styles.root} onClick={handleLoginClick}>
```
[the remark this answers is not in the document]
olehrakwix: fixed

src/components/user-login/user-login.tsx
```diff
@@ -0,0 +1,60 @@
...
+            {isUserLoggedIn ? (
+                <Select
+                    className={styles.select}
+                    dropdownClassName={styles.dropdown}
+                    value={LoginMenuAction.initial}
+                    onValueChange={handleAction}
+                    placeholder={avatar}
+                    renderValue={() => avatar}
```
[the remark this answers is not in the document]
olehrakwix: nice idea, fixed

lib/ecom/api.ts
olehrakwix: refactored as discussed

src/components/header/header.module.scss
olehrakwix: refactored layout

alisey:
I have a lot of questions and comments, reviewing everything in one PR will be unwieldy. We need to split this into at least three PRs:

1. The dropdown menu component (with a board)
2. The user menu UI
3. The backend part

olehrakwix:
@alisey, thanks, this make sense, here are few smaller PRs:
- backend: https://github.com/codux-templates/e-commerce-reclaim/pull/193
- dropdown menu component: https://github.com/codux-templates/e-commerce-reclaim/pull/194
- avatar component: https://github.com/codux-templates/e-commerce-reclaim/pull/195
- header menu and actions refactoring: https://github.com/codux-templates/e-commerce-reclaim/pull/196

one or two more will come once these are merged

lib/hooks/use-member-info.ts
```diff
@@ -0,0 +1,18 @@
...
+export function useMemberInfo() {
+    const api = useEcomApi();
+
+    const [member, setMember] = useState<Member>();
+
+    const isLoggedIn = useMemo(() => api.isUserLoggedIn(), [api]);
+
+    useEffect(() => {
+        if (api.isUserLoggedIn()) {
+            api.getUser().then((u) => setMember(u));
```
VladyslavG, line 13: you can also `api.getUser().then(setMember);`

yurii-ve:
@olehrakwix on the screenshot attached dropdown menu is too close to the dropdown trigger

lib/hooks/use-member-info.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export function useMemberInfo() {
+    const api = useEcomApi();
+
+    const [member, setMember] = useState<Member>();
+
+    const isLoggedIn = useMemo(() => api.isUserLoggedIn(), [api]);
+
+    useEffect(() => {
+        if (api.isUserLoggedIn()) {
```
yurii-ve, line 12: `isLoggedIn`?

lib/hooks/use-member-info.ts
```diff
@@ -0,0 +1,18 @@
+import { useEffect, useMemo, useState } from 'react';
+import { Member, useEcomApi } from '../ecom';
+
+export function useMemberInfo() {
```
yurii-ve, line 4: already mentioned this above, I'd rather use `user` instead of `member` in the code base.

src/components/user-menu/user-menu.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { ChevronDownIcon } from '../icons';
+
+import styles from './user-menu.module.scss';
+
+export const UserMenu = () => {
+    const { isLoggedIn, member } = useMemberInfo();
+
+    const navigate = useNavigate();
+
+    const handleRootClick = () => {
```
yurii-ve, line 18: I believe not using `DropdownMenu` component at all if user is not logged in would simplify this component a lot

```
if (!isLoggedIn) {
    return <div onClick={() => navigate('/login')}>Log In</div>
} else {
    return <DropdownMenu />
}
```

olehrakwix: simplified layout

src/components/user-menu/user-menu.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+import styles from './user-menu.module.scss';
+
+export const UserMenu = () => {
+    const { isLoggedIn, user } = useUserInfo();
+
+    if (!isLoggedIn) {
+        return (
+            <NavLink className={classNames(styles.link, styles.root)} to={'/login'}>
+                <Avatar imageSrc={user?.profile?.photo?.url} />
```
yurii-ve, line 20: isn't this suppose to be anonymous user avatar?

lib/hooks/use-user-info.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export function useUserInfo() {
+    const api = useEcomApi();
+
+    const [user, setUser] = useState<Member>();
+
+    const isLoggedIn = useMemo(() => api.isLoggedIn(), [api]);
+
+    useEffect(() => {
+        if (api.isLoggedIn()) {
```
yurii-ve, line 12: `isLoggedIn`? (constant above)

I'll aks again if you'll ignore it again 😄

src/components/icons/chevron-down-icon.tsx
```diff
@@ -0,0 +1,9 @@
+export const ChevronDownIcon = (props: React.SVGProps<SVGSVGElement>) => (
```
yurii-ve, line 1: can't the `DropdownIcon` be used?

src/components/user-menu/user-menu.tsx
```diff
@@ -0,0 +1,60 @@
...
+            </NavLink>
+        );
+    }
+
+    return (
+        <DropdownMenu
+            trigger={
+                <button className={styles.root}>
+                    <Avatar imageSrc={user?.profile?.photo?.url} />
+                    <ChevronDownIcon width={10} height={10} />
```
yurii-ve, line 31: we have `DropdownIcon`, can't it be used here?

src/components/user-menu/user-menu.tsx
olehrakwix: nice catch, missed this while refactoring...
lets keep `Avatar` component here, it handles `undefined` value correctly

lib/hooks/use-user-info.ts
olehrakwix: didn't get what you mean previously😀

src/components/user-menu/user-menu.tsx
olehrakwix: you are right, removed that new icon and used `DropdownIcon` here

yurii-ve (approved):

PR #179, opened 2024-10-30
feat: remove fallback to all products in featured products section

alisey:
Since we now have skeletons in the `<FeaturedProductsSection>`, showing them if the category doesn’t exist seems better than falling back on all products. This makes it easier for users to understand that something isn't working, while still giving them an idea of what this section should display.

This change also allowed to remove `.getFeaturedProducts()` from the `EcomAPI`, simplify the `FeaturedProductsSection` component by removing `useSWR` from it, and reduce the overall amount of code a little.

---

Skeletons:

Screenshot 2024-10-30 at 14 35 17

### DIFF
--- a/lib/ecom/hooks.ts
+++ b/lib/ecom/hooks.ts
@@ -1,9 +1,9 @@
 import { useEffect, useState } from 'react';
-import useSwr, { Key } from 'swr';
+import useSwr, { Key, SWRResponse } from 'swr';
 import useSWRMutation from 'swr/mutation';
 import { findItemIdInCart } from '~/lib/utils';
 import { useEcomAPI } from './api-context';
-import { AddToCartOptions } from './types';
+import { AddToCartOptions, CollectionDetails, GetProductsOptions, Product } from './types';
 
 export const useCartData = () => {
     const ecomApi = useEcomAPI();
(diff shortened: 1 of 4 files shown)


lib/ecom/hooks.ts
```diff
@@ -169,3 +169,39 @@ export const useCart = () => {
...
+    const ecomApi = useEcomAPI();
+    return useSwr(
+        ['category-details', slug],
+        async ([, slug]) => {
+            const response = await ecomApi.getCategoryBySlug(slug);
+            if (response.status === 'failure') throw response.error;
+            return response.body;
+        },
+        {
+            keepPreviousData: false,
```
alisey, line 183: `keepPreviousData: false` - allows us to switch to skeletons if the user changes the `categorySlug` via Codux to something invalid. Otherwise the component would display the last valid response from the old slug.

src/components/featured-products-section/featured-products-section.tsx
```diff
@@ -17,32 +16,22 @@ interface FeaturedProductsSectionProps {
...
-    );
+    const category = useCategoryDetails(categorySlug);
+    const products = useProducts({ categorySlug, limit: productCount });
 
     return (
         <div className={classNames(styles.root, className)}>
             <FadeIn className={styles.header} duration={1.8}>
-                <h3 className={styles.headerTitle}>{title ?? data?.category.name}</h3>
+                <h3 className={styles.headerTitle}>
+                    {title ?? category.data?.name ?? categorySlug}
```
alisey, line 26: Added `categorySlug` as a fallback title.

tetianaman (approved):

PR #180, opened 2024-10-30
refactor(products): separate product list to own component

AviVahl:
- doing this for upcoming search route, which should reuse most of this route's UI
- turned off eslint rule (react/prop-types) that is irrelevant for typescript react projects


src/components/product-list/product-list.module.scss
```diff
@@ -0,0 +1,19 @@
+.productsList {
```
alisey, line 1: Let's rename to `.root` or `.productList` (singular)

alisey (approved):
Not sure if "ProductList" is the best name, given that it may render a message referring to "Try different filters or another category" when intuitively I would expect it to simply return an empty list.

I don’t have a better alternative off the top of my head.

src/components/product-list/product-list.module.scss
AviVahl: renamed to `productList`. `root` feels too stylable-specific.

AviVahl:
> I don’t have a better alternative off the top of my head.

I thought about that, but I have no better name as well. I expect further refactors going forward (for the search), as we need the filtering ui as well to be reused.

How about `ProductGrid`? I expect this component to not include the empty template/no results parts in the future.

src/components/product-list/product-list.module.scss
AviVahl: also moved a missed media query (verified the other classNames as well now)

alisey:
`ProductGrid` sounds fine.

PR #183, opened 2024-10-31
feat: my account and my orders links

olehrakwix:
image


src/components/user-login/user-login.module.scss
```diff
@@ -21,12 +21,15 @@
 
 body .dropdown {
```
yurii-ve, line 22: why `body` is here?

olehrakwix: without it for some reasons my class does not override default one (even though my class comes second in a className list). and I can not define any other child except `body`, because drop down is opened in portal

olehrakwix: just checked again and now it works as expected... not sure why this happened, removing `body`

src/components/user-login/user-login.tsx
```diff
@@ -42,6 +54,21 @@ export const UserLogin = () => {
                     placeholder={avatar}
                     renderValue={() => avatar}
                 >
+                    <SelectItem
```
yurii-ve, line 57: using `Select` component for this purpose looks weird to me.. and quite confusing to read

yurii-ve: I wouldn't expect that editing `Select` component will result in change to the user menu in the header

PR #185, opened 2024-11-01
feat: add meta tags

yurii-ve:
- root page: title, description, robots, og tags and twitter tags with static cover image (added to the `/public`)
- category page: dynamic title and description, robots
- product details page: dynamic title and description, robots, og tags and twitter tags with dynamic cover image
- other pages: only title, description and robots

Note: OG tags and twitter tags are added to make the shared via social media or messenger link look nice (see screenshot with shared e-commerce-remix link below)

image


app/root.tsx
```diff
@@ -133,3 +129,49 @@ export function ErrorBoundary() {
         </ContentWrapper>
     );
 }
+
+export const meta: MetaFunction = () => {
+    const title = 'ReClaim: Home Goods Store';
+    const description = 'Essential home products for sustainable living';
+    const imagePath = '/cover.jpg';
```
alisey, line 136: How about `open-graph-image.jpg` or `social-media-image.jpg` instead of nondescript `cover.jog`?

app/routes/_index/route.tsx
```diff
@@ -97,3 +98,49 @@ export default function HomePage() {
         </div>
     );
 }
+
+export const meta: MetaFunction = () => {
```
alisey, line 102: Doesn't this automatically get inherited from `root`? Why does it have to be repeated here?

alisey: Yikes. They say here that they don't do merging:
https://remix.run/docs/en/main/route/meta

So if a route returns `meta`, then only that `meta` will be used. And if it doesn't return `meta`, then parent's `meta` will be used.

alisey: Do we need these things to defined statically for the ability to edit them through Codux?

If not, we could create a helper to reduce boilerplate.
```ts
export const meta = createMeta({
    title: '...',
    description: '...',
    image: '...'
});
```

app/routes/product-details.$productSlug/route.tsx
```diff
@@ -204,3 +210,49 @@ export function ErrorBoundary() {
         />
     );
 }
+
+export const meta: MetaFunction<typeof loader> = ({ data }) => {
+    const title = data?.product.name ?? 'Product Details';
+    const description = data?.product.description ?? 'Product Description';
```
alisey, line 216: It's better to omit it entirely than to add "Product Description" as the description.

app/routes/products.$categorySlug/route.tsx
```diff
@@ -236,3 +242,17 @@ export function ErrorBoundary() {
         />
     );
 }
+
+export const meta: MetaFunction<typeof loader> = ({ data }) => {
+    return [
+        { title: data?.category.name ?? 'ReClaim: Products' },
+        {
+            name: 'description',
+            content: data?.category.description ?? 'Category description',
```
alisey, line 251: Same here, better leave it empty than "Category description".

app/routes/_index/route.tsx
yurii-ve: As far as I know right now it's not possible to edit them through Codux, but I stillI prefer having them set explicitly. so the user can easily set additional meta tags through code in a way it's described in Remix docs.

app/routes/_index/route.tsx
```diff
@@ -97,3 +98,32 @@ export default function HomePage() {
...
```
alisey, line 102: This one can prolly be omitted since it's identical to the global.

alisey (approved):

PR #186, opened 2024-11-01
refactor: migrate the 'Thank You' page to exceptions

alisey:


.eslintrc.cjs
```diff
@@ -24,6 +24,14 @@ module.exports = {
...
 
     overrides: [
+        // General
+        {
+            files: ['**/*.{ts,tsx}'],
+            rules: {
+                'no-empty': 'off',
+            },
+        },
+
```
alisey, line 34: Note: this allows empty catch blocks without a pointless comment.

Before:
```ts
try { JSON.parse() } catch { /* duh */ }
```

After:
```ts
try { JSON.parse() } catch { }
```

app/routes/thank-you/route.tsx
```diff
@@ -10,9 +9,12 @@ import { OrderSummary } from '~/src/components/order-summary/order-summary';
 import styles from './route.module.scss';
 
 export const loader = async ({ request }: LoaderFunctionArgs) => {
+    const url = new URL(request.url);
+    const orderId = url.searchParams.get('orderId');
     const api = await initializeEcomApiForRequest(request);
-
-    return getThankYouRouteData(api, request.url);
+    // Allow a missing `orderId` to support viewing this page in Codux.
+    const order = orderId ? await api.getOrder(orderId) : undefined;
```
alisey, line 16: Note: inlined the `getThankYouRouteData` loader since there's no business logic to abstract here.

`api.getOrder(orderId)` returns undefined if the order is not found. I think it's better to show the "Thank You" page without the order details then to show the "Order Not Found" page.

app/routes/thank-you/route.tsx
```diff
@@ -39,7 +41,5 @@ export default function ThankYouPage() {
 
 export function ErrorBoundary() {
     const error = useRouteError();
-    const title = isRouteErrorResponse(error) ? 'Failed to load order details' : 'Error';
-    const message = getErrorMessage(error);
-    return <ErrorPage title={title} message={message} />;
+    return <ErrorPage title="Error" message={getErrorMessage(error)} />;
```
alisey, line 44: Note: this will show a more specific error subtitle instead of "Failed to load order details":

image

lib/ecom/api.ts
```diff
@@ -243,17 +249,10 @@ function createEcomApi(wixClient: WixApiClient): EcomAPI {
...
         async getOrder(id) {
             try {
-                const order = await wixClient.orders.getOrder(id);
-                if (!order) {
-                    return failureResponse(EcomApiErrorCodes.OrderNotFound, 'Order not found');
-                }
-
-                return successResponse(order);
-            } catch (e) {
-                if (isEcomSDKError(e) && e.details.applicationError.code === 404) {
```
alisey, line 253: > `if (isEcomSDKError(e) && e.details.applicationError.code === 404)`

Note: this condition was incorrect, `WixClient` returns `"NOT_FOUND"` instead of `404` for this method 🤦‍♂️

lib/utils/common.ts
```diff
@@ -27,28 +26,25 @@ export function removeQueryStringFromUrl(url: string) {
...
-
     // Remix ErrorResponse thrown from an action or loader:
-    // - throw new Response('oops');
-    // - throw json('oops')
-    // - throw json({message: 'oops'})
+    // - throw new Response('Oops', { status: 404 })
+    // - throw json({message: 'Oops'}, { status: 404 })
     if (isRouteErrorResponse(error)) {
-        error = error.data;
+        if (error.data) {
```
alisey, line 41: Note: this is for better handing of:
```ts
throw new Response(null, { status: 500, statusText: 'Oops'})
```
Where `error.data` would be `null`.

app/routes/thank-you/route.tsx
olehrakwix: don't we want to show both? e.g.
```
Failed to load order details:
getaddrinfo ENOTFOUND www.wixapis.com
```

lib/ecom/wix-client-error.ts
olehrakwix, line 1: some crazy staff... just no time to explain 😀

olehrakwix (approved):

app/routes/thank-you/route.tsx
alisey: It would be a little bit nicer. We could do:

```ts
    try {
        const order = orderId ? await api.getOrder(orderId) : undefined;
        return { order };
    } catch (error) {
        throw json({ message: `Failed to load order. ${getErrorMessage(error)}`}, { status: 500 });
    }
```

Or:
```ts
throw unexpectedErrorResponse('Failed to load order.', error);
```

Because of how unlikely this error is, I'd prefer to keep things simple and just let the original error bubble up.

PR #188, opened 2024-11-01
refactor: automatically normalize WixClient errors

alisey:
`WixClient` throws errors with garbled messages, to fix them we have to chain `.catch(throwNormalizedWixClientError)` to each method call of `WixClient` inside of `EcomApi`. This is error-prone.

This PR introduces a helper function that iterates over all methods of `EcomApi` and wraps them with `normalizeWixClientError` automatically. The code is hacky, but it makes `EcomApi` methods nice and clean:

```ts
getCart() {
    return wixClient.currentCart.getCurrentCart();
}
```

Other alternatives:

1. Add a try-catch inside of each `EcomApi` method:
```ts
async getCart() {
    try {
        return await wixClient.currentCart.getCurrentCart();
    } catch (error) {
        throw normalizeWixClientError(error);
    }
}
```

2. Wrap each `EcomApi` method individually:
```ts
getCart: withNormalizedWixClientErrors(() => {
    return wixClient.currentCart.getCurrentCart();
})
```

3. Use experimental decorators (this will require converting `createEcomApi` to a class):
```ts
@normalizeWixClientErrors
getCart() {
   return wixClient.currentCart.getCurrentCart();
}
```

Let me know if you prefer one of the alternatives. The downside is that they require adding something to each method manually, and failing to do so won't have an obvious effect.

### DIFF
--- a/lib/ecom/api.ts
+++ b/lib/ecom/api.ts
@@ -12,7 +12,7 @@ import {
     EcomApiFailureResponse,
     EcomApiSuccessResponse,
 } from './types';
-import { isNotFoundWixClientError, throwNormalizedWixClientError } from './wix-client-error';
+import { isNotFoundWixClientError, normalizeWixClientError } from './wix-client-error';
 
 type WixApiClient = WixClient<
     undefined,
@@ -292,10 +292,12 @@ const withNormalizedWixClientErrors = (api: EcomApi): EcomApi => {
             try {
                 const result = Reflect.apply(original, api, args);
                 if (result instanceof Promise) {
-                    return result.catch(throwNormalizedWixClientError);
+                    return result.catch((error) => {
+                        throw normalizeWixClientError(error);
+                    });
                 }
             } catch (error) {
-                throwNormalizedWixClientError(error);
+                throw normalizeWixClientError(error);
             }
         });
     }
--- a/lib/ecom/api.ts
+++ b/lib/ecom/api.ts
@@ -68,14 +68,12 @@ export function initializeEcomApiAnonymous() {
     return createEcomApi(client);
 }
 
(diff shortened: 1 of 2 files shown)


lib/ecom/api.ts
```diff
@@ -285,8 +278,32 @@ function createEcomApi(wixClient: WixApiClient): EcomApi {
...
         },
-    };
-}
+    });
+
+/**
+ * Wraps all methods of the EcomApi with a try-catch block that fixes broken
+ * error messages in WixClient errors and rethrows them.
+ */
+const withNormalizedWixClientErrors = (api: EcomApi): EcomApi => {
```
olehrakwix, line 287: I'd recomment to move this to a separate file. I understand that for now it is not needed, but I that API growth, so I can image we will decide to split it to few smaller sub-APIs. in that case this will help

olehrakwix (approved):

lib/ecom/api.ts
alisey: Since this function mutates the original object, I only feel comfortable defining it next to the object it mutates. Using this function as a black box without understanding its implementation could lead to tricky bugs.

For example, if EcomApi was defined as a class:
```ts
class EcomApi {
    getCart() { /* ... */ }
}
```

TypeScript would happily let you do:
```ts
const api = withNormalizedWixClientErrors(new EcomApi());
```

But later you would discover that error normalization has stopped working.

To solve the issues with mutation and non-iterable properties I could use a proxy, but it has more subtle issues, such `ecomApi.getOrder` not being a stable reference and returning a new function each time you access it.

If we're going to split `EcomApi` into multiple files, I think one of the alternative approaches outlined in the PR description would be safer and more predictable.

PR #193, opened 2024-11-03
feat: user login backend

olehrakwix:
added routes and logic for user login

- `/login` - redirects user to a wix managed login page
- `/logout` - logs user out from wix managed login system and clears session
- `/login-callback` - handles callback from wix managed login system to obtain logged in member tokens, saves it to a session and redirects user to a specified in search params page


app/routes/login-callback/route.tsx
```diff
@@ -0,0 +1,37 @@
+import { LoaderFunctionArgs, redirect } from '@remix-run/node';
+import { commitSession, getSession, initializeEcomApiForRequest } from '~/lib/ecom/session';
+
+export async function loader({ request }: LoaderFunctionArgs) {
+    const session = await getSession(request.headers.get('Cookie'));
+    const storedOauthData = session.get('oAuthData');
+
+    if (storedOauthData === undefined) {
+        return redirect('/login');
```
yurii-ve, line 9: I know that Codux won't show this route, but let's still double check that everything works fine in the define app mode, as it doesn't support node-specific code.

app/routes/login-callback/route.tsx
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    const api = await initializeEcomApiForRequest(request);
+    const wixClient = api.getWixClient();
+
+    const returnedOAuthData = wixClient.auth.parseFromUrl(request.url, 'query');
+    if (returnedOAuthData.error) {
+        throw new Error(`Error: ${returnedOAuthData.errorDescription}`);
```
yurii-ve, line 17: will the root error boundary catch this and show nice error UI?

app/routes/login-callback/route.tsx
```diff
@@ -0,0 +1,37 @@
...
+    const memberTokens = await wixClient.auth.getMemberTokens(
+        returnedOAuthData.code,
+        returnedOAuthData.state,
+        storedOauthData,
+    );
+    wixClient.auth.setTokens(memberTokens);
+
+    session.set('wixEcomTokens', memberTokens);
+
+    const redirectUrl = new URL(storedOauthData.originalUri || new URL(request.url).host);
```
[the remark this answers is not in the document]
olehrakwix: we are creating a new instance of `URL` class here. if value is empty string - it will fail with errror, so I'd like to have `||` in this case
image

app/routes/login-callback/route.tsx
olehrakwix: no, it will look like this:
image

olehrakwix: no, it will look like this:
image

yurii-ve: not cool

olehrakwix: I can see two options to make it better:
1. create a component and error boundary for this route (not desired, cause that page will be visible in codux and basically it is not clear what to there - spinner?)
2. rediret to some `/error` route, but this is also not good, we were trying to omit separate error page

olehrakwix: might look like this, but in this case, page appear in codux:
image

app/routes/login-callback/route.tsx
```diff
@@ -0,0 +1,37 @@
...
+    redirectUrl.searchParams.set('login', 'success');
```
alisey, line 30: > redirectUrl.searchParams.set('login', 'success');

What is this for?

app/routes/login-callback/route.tsx
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    const memberTokens = await wixClient.auth.getMemberTokens(
+        returnedOAuthData.code,
+        returnedOAuthData.state,
+        storedOauthData,
+    );
+    wixClient.auth.setTokens(memberTokens);
+
+    session.set('wixEcomTokens', memberTokens);
```
alisey, line 27: Are these actually ecom tokens, or Wix API tokens in general?

alisey: No, it's not ecom tokens.
```ts
wixEcomTokens = await client.auth.generateVisitorTokens();
```

app/routes/login-callback/route.tsx
olehrakwix: this is done because netlify intentionally left original query parameters on redirects.
the flow:
1. user goes to `/login` - login process is initiated and user is redirected to wix login page
2. user logs in on wix login page
3. wix login page redirects user to `/login-callback?code=<oAuth login code>&state=<oAuth login state>`
4. our loader finishes login process and redirect user to original page (from where login was initiated) with 302 response code and locaiton like `/categories/all-products`
5. netlify keeps query params and user is redirected to `/categories/all-products?code=<oAuth login code>&state=<oAuth login state>`

in order to avoid `code` and `state` presence in URL there are two options:
1. `useEffect` and clear search params
2. add own search param when redirecting

we've discussed this with @yurii-ve  and decided it is better to add dummy `login=success` parameter when redirecting instead of clearing `code` and `state` params.
also, we haven't added a comment about it in code, cause we wanted to omit any netlify mentions, as it is not directly related to a template itself

link to netlify support answer that this is expected and done intentionally:
https://answers.netlify.com/t/netlify-keeps-searchparams-when-redirecting-in-my-next-js-app/108590/2

app/routes/login-callback/route.tsx
olehrakwix: yeah, I agree. what name do you like more, `wixTokens`, `wixAuthTokens`, or something else?

app/routes/login-callback/route.tsx
alisey: Things like this absolutely must have a comment in code. This code looks unused and I could have easily deleted it during a refactor.

app/routes/login-callback/route.tsx
alisey: `wixVisitorTokens`

alisey: If I understand correctly what they are.

> https://dev.wix.com/docs/sdk/core-modules/sdk/oauth-strategy#generatevisitortokens

They say: use this function when creating a session for visitors.

"Visitors" includes not logged in visitors. But if these tokens are not simply session tokens, but authentication tokens, then `wixAuthTokens`.

olehrakwix: they call anonymous users "visitors" and logged in users "members"

lib/ecom/api.ts
```diff
@@ -180,6 +171,19 @@ const createEcomApi = (wixClient: WixApiClient): EcomApi =>
...
+
+            const { authUrl } = await wixClient.auth.getAuthUrl(oAuthData, {
+                responseMode: 'query',
+            });
+
+            return { oAuthData, authUrl };
+        },
+        async logout(returnUrl: string) {
+            const result = await wixClient.auth.logout(returnUrl);
+            return result;
```
alisey, line 185: `return wixClient.auth.logout(returnUrl);`

app/routes/logout/route.tsx
```diff
@@ -0,0 +1,17 @@
+import { LoaderFunctionArgs, redirect } from '@remix-run/node';
+import { destroySession, getSession, initializeEcomApiForRequest } from '~/lib/ecom/session';
+
+export async function loader({ request }: LoaderFunctionArgs) {
+    const currentUrl = new URL(request.url);
+    const returnUrlPath = currentUrl.searchParams.get('returnPath') || '/';
```
alisey, line 6: Who's going to set `returnPath`?

app/routes/login/route.tsx
```diff
@@ -0,0 +1,23 @@
+import { LoaderFunctionArgs, redirect } from '@remix-run/node';
+import { commitSession, getSession, initializeEcomApiForRequest } from '~/lib/ecom/session';
+
+export async function loader({ request }: LoaderFunctionArgs) {
+    const currentUrl = new URL(request.url);
+    const returnUrlPath = currentUrl.searchParams.get('returnPath') ?? '/';
```
alisey, line 6: Why don't we always return them to the page from which they've logged in?

olehrakwix: they not always log in from home page. they might click login from any other public page, e.g. category, product details. it is better to redirect back to same page.
returnPath is set in query params when user click login button

app/routes/login-callback/route.tsx
alisey: Do we need to have these tokens for anonymous visitors as well?

olehrakwix: correct. because anonymous user may add products to a cart and even place an order. it is not required to log in for that. so, the session is generated for anonymous users as well.
from wix api we have `generateVisitorTokens` to create new tokens for anonymous user and `getMemberTokens` that finishes login process and generates tokens for logged in user

app/routes/login-callback/route.tsx
alisey: > cause we wanted to omit any netlify mentions

We won't explain what this code is for, because we don't talk about Ne*lify around here.

app/routes/login-callback/route.tsx
alisey: Then `wixVisitorTokens` sounds like an appropriate name.

olehrakwix: I like `wixAuthTokens` better, WDYT?

app/routes/logout/route.tsx
olehrakwix: there is another comment about login and `returnPath`, but for logout I agree that maybe it makes sense to always redirect to home page

app/routes/login/route.tsx
alisey: This is exactly what I'm saying. Why don't we always redirect them to where they came from instead of manually setting return path? Do we not have the referrer header?

app/routes/login-callback/route.tsx
alisey: Didn't we just discussed that these are visitor tokens and not authentication tokens.

alisey: Do we save them in our session under the same key both for logged in and anonymous users?

alisey: Yes we do:

https://github.com/codux-templates/e-commerce-reclaim/blob/0d39cc8388d1913989e47b0e3c6fca81bec11670/lib/ecom/session.ts#L43-L44

So if you name them `wixAuthTokens` how does it make sense to set them for anonymous visitor?

alisey: We also need to explain why we save `wixClientId` into the session. Because normally that doesn't make sense, it's not session-dependent information.

olehrakwix: what is a difference between `visitor` and `authentication` tokens in your eyes?
I mean, these tokens authenticates particular visitor for wix services

olehrakwix: https://dev.wix.com/docs/go-headless/coding/java-script-sdk/visitors-and-members/handle-visitors

> Once you've [created a client](https://dev.wix.com/docs/go-headless/coding/java-script-sdk/visitors-and-members/create-a-client-with-oauth), your client can generate, manage, and use tokens for anonymous site visitors. The client uses these tokens when making requests to Wix APIs on behalf of a visitor to maintain the visitor's session.


so, anonymus user is just an unknown user, but it has session and authorization

yurii-ve: > Then wixVisitorTokens sounds like an appropriate name.

@alisey anyone who will go to wix sdk docs will be confused, because docs say that visitor is an anonymous user, but code says that `wixVisitorTokens` is used for both anonymous and logged in users. 
maybe `wixSessionToken`?

app/routes/login-callback/route.tsx
yurii-ve: @olehrakwix sorry, probably removing this comment wasn't among the best of my ideas 😅

app/routes/login/route.tsx
olehrakwix: this is exactly what we are doing here. but idea with referrer is better, thanks, will change this

app/routes/login-callback/route.tsx
olehrakwix: image

olehrakwix: added comment

app/routes/login-callback/route.tsx
olehrakwix: renamed to `wixSessionTokens `

app/routes/login/route.tsx
olehrakwix: switch to `referer` header

app/routes/logout/route.tsx
olehrakwix: redirect to home page after logout

app/routes/login-callback/route.tsx
olehrakwix: >We also need to explain why we save wixClientId into the session. Because normally that wouldn't make sense, it's not session-dependent information.

there already was a comment few lines above where we reading it. anyway, improved that comment, added one more at the line where we saving it to session

yurii-ve (approved):

PR #194, opened 2024-11-03
feat: add dropdown menu component

olehrakwix:
add dropdown menu component. this component will be used for user/login menu

image

image

### DIFF
--- a/src/components/dropdown-menu/dropdown-menu.tsx
+++ b/src/components/dropdown-menu/dropdown-menu.tsx
@@ -5,11 +5,11 @@ import React from 'react';
 import styles from './dropdown-menu.module.scss';
 
 export interface DropdownMenuProps extends React.PropsWithChildren {
-    trigger: React.ReactNode;
+    trigger: React.ReactElement;
     contentProps?: RadixDropdownMenu.DropdownMenuContentProps & React.RefAttributes<HTMLDivElement>;
 }
 
-export const DropdownMenu = ({ trigger, children, contentProps }: DropdownMenuProps) => (
+export const DropdownMenu = ({ trigger, children, contentProps = {} }: DropdownMenuProps) => (
     <RadixDropdownMenu.Root>
         <RadixDropdownMenu.Trigger asChild className={styles.trigger}>
             {trigger}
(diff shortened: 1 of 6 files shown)


alisey:
Pushed several tweaks:

1. Used `--radix-dropdown-menu-content-available-width` and `--radix-dropdown-menu-content-available-height` to ensure that the menu doesn't extend outside the viewport. Added a vertical scroll in case it needs one.

2. Rearranged some styling, removed `.contentWrapper` which sounded like a wrapper of the `.content` element rather than a wrapper of the menu item content.

3. Made the item focus effect more pronounced (it doesn't match the reference design, but it's actually visible now). Changed it from `:focus-visible` to `:focus, :hover` since sometimes Radix doesn't add focus state to hovered elements.

4. Changed the `trigger` type to `ReactElement` instead of `ReactNode`, since it's rendered `asChild` and Radix cannot render plain text nodes with that attribute since it's indicates that props are passed to the child, and text nodes don't accept props.

5. Added a comment explaining `onCloseAutoFocus={(e) => e.preventDefault()}`

6. Removed redundant property spreading from `MenuItem`.

Screenshot 2024-11-05 at 10 10 13

alisey:
In Radix docs they mention that components must forward ref:
https://www.radix-ui.com/primitives/docs/guides/composition#your-component-must-forward-ref

I can't say that I fully understand why and when. everything seems to work fine here without it.

src/components/dropdown-menu/dropdown-menu.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+export type DropdownMenuItemProps = RadixDropdownMenu.DropdownMenuItemProps &
+    React.RefAttributes<HTMLDivElement>;
+
+export const DropdownMenuItem = ({ className, ...restProps }: DropdownMenuItemProps) => (
+    <RadixDropdownMenu.Item className={classNames(styles.item, className)} {...restProps} />
+);
+
+export type DropdownMenuSeparatorProps = RadixDropdownMenu.DropdownMenuSeparatorProps &
+    React.RefAttributes<HTMLDivElement>;
```
alisey, line 39: > React.RefAttributes<HTMLDivElement>;

@olehrakwix you've added this, but you're not actually passing a ref? I know in React 19 or something they want to remove `forwardRef`, but we currently use React 18 and it doesn't allow to pass ref via props.

olehrakwix:
> In Radix docs they mention that components must forward ref:
https://www.radix-ui.com/primitives/docs/guides/composition#your-component-must-forward-ref
> I can't say that I fully understand why and when. Everything seems to work fine here without it.

the same. I didn't see this in documentation, but component actually works without any issues.
also, `RadixDropdownMenu.Root` does not accept `ref` property. `RadixDropdownMenu.Content` and `RadixDropdownMenu.Trigger` does, but it is not clear where should we forward ref from root component

PR #196, opened 2024-11-03
refactor: get rid of empty div to align header menu and actions

olehrakwix:
get rid of empty div and its fixed width to properly align menu and actions

image

image


tetianaman:
Hi @olehrakwix !

Check if in case of many nav items the most right one does not overlap LogIn button(link)
Screenshot 2024-11-04 at 09 17 45

olehrakwix:
@tetianaman , so what is expected result if items physically don't have enough space?
because at the moment whole page horizontal scroll appear, and this also looks weird

tetianaman:
> what is expected result if items physically don't have enough space? 

Looking at the Wix Studio template we have items wrapping

Screenshot 2024-11-04 at 10 00 58

I think it's ok, but lets keep LogIn, Cart (and Menu burger icon) at the top

olehrakwix:
@tetianaman , done

tetianaman:
The second image in PR description has a bit big indent between cart and burger menu icons, but I don't see it running the project.

alisey:
> Looking at the Wix Studio template we have items wrapping

@tetianaman and we have a variable that defines header height, which is incompatible with wrapping.

tetianaman:
@alisey 
Right... 

> define hero banner height using viewport units

In that case it will look cut at the bottom. Are we okay with it?

Screenshot 2024-11-04 at 14 56 18

alisey:
I'm OK with the hero banner being a little cut off. The header height variable would cause problems for users anyway because it's difficult to keep it up to date (this PR is a testament to that).

@tetianaman please remove all variables related to header height.

@olehrakwix if the header is going to wrap to multiple lines, let's keep the user menu and the cart aligned to the top, and have more space between menu and the image here:

image

Let's use same distance as between menu rows.

src/components/header/header.module.scss
```diff
@@ -36,7 +29,8 @@
     display: grid;
     grid-template-columns: 1fr auto 1fr;
     align-items: center;
-    min-height: var(--headerNavigationHeight);
+    justify-content: center;
```
tetianaman: removed in the next commit

src/components/header/header.module.scss
```diff
@@ -1,18 +1,11 @@
...
-}
-
 .root {
     padding: 0 var(--pagePaddingHoriz);
-    min-height: var(--headerHeight);
 }
 
 .topBar {
-    height: var(--headerTopBarHeight);
+    height: 50px;
```
alisey, line 8: Why does it need explicit height rather than being based on content?

src/components/header/header.module.scss
```diff
@@ -36,7 +29,8 @@
     display: grid;
     grid-template-columns: 1fr auto 1fr;
     align-items: center;
-    min-height: var(--headerNavigationHeight);
+    justify-content: center;
+    min-height: 54px;
```
alisey, line 32: Why does it need min-height.

olehrakwix: fixed

tetianaman (approved):

alisey (approved):

PR #201, opened 2024-11-04
feat: add text input component styles

olehrakwix:
image

image


_codux/ui-kit/components/components.board.tsx
```diff
@@ -17,6 +17,25 @@ export default createBoard({
         <ComponentWrapper>
             <Kit category="Core Components" title="Components & Elements">
                 <Kit.Section title="Inputs">
+                    <Kit.Item>
+                        <Variant name="Input">
+                            <input className="text-input" value="text input" onChange={() => {}} />
```
yurii-ve, line 22: I'm thinking if we should provide `value` and `onChangeProps` as it means that user won't be able to type anything in the board.
On the other hand these are convenient to have when user drops this input from the add panel.

@alisey what do you think?

src/styles/common.scss
```diff
@@ -90,6 +90,28 @@
...
+    background: transparent;
+    padding: 8px 0;
+    border-bottom: solid 1px rgba(23, 17, 13, 0.4);
+}
+
+input.text-input {
+    box-shadow: none;
+}
+
+input.text-input:disabled {
```
yurii-ve, line 106: how does the placeholder text look like? maybe it should be styled it as well?

olehrakwix: in wix template there is no placeholders:
Screenshot 2024-11-05 at 12 56 33

so I'm not sure what is expected in this case

olehrakwix: anyway, added some styles for placeholder and added this state to a board

yurii-ve (approved):

src/styles/common.scss
```diff
@@ -113,6 +113,29 @@
     color: inherit;
 }
 
+input.text-input {
```
alisey, line 116: Don't use kebab case.

PR #202, opened 2024-11-05
refactor: replace css reset

idoros:
This PR replaces the existing aggressive CSS reset with the [Josh Comeau CSS reset](https://www.joshwcomeau.com/css/custom-css-reset/), aiming to preserve the basic styling of HTML elements when unstyled.

### Changes in this PR:
- Replaced the current CSS reset with a simpler approach to maintain default element styles.
- Moved common CSS to load first, ensuring it does not override styles with equal specificity.
- Added `transparentInput` CSS variant to serve as base for embedded input in custom controllers (`quantity-input`).
- Added `iconButton` and `linkButton` CSS variants
- Set a stacking context for the site wrapper.
- Refactored components to work with the new reset.

### Tech-debt
- Consider unifying base button styles under a single button component.


tomrav:
Spacing above the breadcrumbs on the category page is different to original template.
image

tomrav:
Price filter title looks different
image

tomrav:
very small diff in the title spacing on privacy policy (and refund, shipping, and terms & conditions)
image

tomrav:
difference in strip size, might be because of the title size
image

tomrav:
input focus on product details looks different
image

tomrav:
Shop Now padding is different
image

app/routes/privacy-policy/route.module.scss
```diff
@@ -0,0 +1,3 @@
+.heading2 {
+    font-weight: 400;
+}
```
alisey, line 3: Why here rather than than in `.textPage h2`?

app/routes/products.$categorySlug/route.module.scss
```diff
@@ -28,6 +28,11 @@
     margin-bottom: 12px;
 }
 
+.categoryList {
+    list-style: none;
+    padding: 0;
+}
```
alisey, line 34: What about margin? `UL` elements also have margin by default.

app/routes/privacy-policy/route.module.scss
idoros: No particular reason, I will change

app/routes/products.$categorySlug/route.module.scss
idoros: The new reset takes care of margin

alisey:
The color select seems to be broken:

image

lib/components/color-select/color-select.module.scss
```diff
@@ -19,7 +19,6 @@
     border: 1px solid var(--border-color);
     border-radius: 50%;
     transition: padding 100ms;
-    cursor: pointer;
```
alisey, line 22: These options are button elements. Buttons don't have `cursor: pointer` by default.

src/styles/reset.scss
```diff
@@ -1,116 +1,64 @@
...
-/* Safari - solving issue when using user-select:none on the <body> text input doesn't working */
+/* 2. Remove default margin */
+* {
+    margin: 0;
+}
+body {
+    /* 3. Add accessible line-height */
+    line-height: 1.5;
+    /* 4. Improve text rendering */
+    -webkit-font-smoothing: antialiased;
```
alisey, line 15: > `-webkit-font-smoothing: antialiased;`

If it's defined here, let's remove it from `index.scss`

src/styles/reset.scss
```diff
@@ -1,116 +1,64 @@
...
+}
+/* 9. Create a root stacking context */
+#root {
+    isolation: isolate;
 }
 
-/* Remove details summary webkit styles */
-::-webkit-details-marker {
-    display: none;
+/* ToDo: addition to josh reset */
```
alisey, line 60: Why `ToDo`?

"Josh Reset" - let's either clearly mark where Josh's reset begins and ends, and add a reference link for those who don't know Josh merely by his name.

src/styles/reset.scss
```diff
@@ -1,116 +1,64 @@
...
+h1,
+h2,
+h3,
+h4,
+h5,
+h6 {
+    text-wrap: balance;
+}
+/* 9. Create a root stacking context */
+#root {
```
alisey, line 56: Shouldn't this be in the index.scss? This an application-specific element ID and styling.

src/components/applied-product-filters/applied-product-filters.module.scss
```diff
@@ -7,7 +7,7 @@
 
 .clearAllButton {
     margin-left: 8px;
+    padding: 0;
     font: var(--paragraph3);
     text-decoration: underline;
-    cursor: pointer;
```
alisey, line 12: Needs cursor.

src/components/cart/cart-item-options/cart-item-options.module.scss
```diff
@@ -9,7 +9,6 @@
     align-items: center;
     gap: 6px;
     font: var(--paragraph3);
-    cursor: pointer;
```
alisey, line 12: I haven't verified everything, but there's probably the same issue with cursor removal everywhere.

src/components/header/header.module.scss
```diff
@@ -46,6 +46,9 @@
     gap: 30px;
 }
 
+.cartBtn {
```
alisey, line 49: `cartButton`

src/styles/reset.scss
idoros: Originally came with the reset, we prefer to keep it here for now.

src/components/product-grid/product-grid.module.scss
```diff
@@ -13,10 +13,10 @@
 
 .clearFiltersButton {
     margin-top: 32px;
-    cursor: pointer;
+    padding: 0;
 }
 
-.clearFiltersButton:hover {
+.clearFiltersButton:not(:disabled):hover {
```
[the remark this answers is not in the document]
idoros: This is because of specificity that is coming from the common variants

src/styles/common.scss
```diff
@@ -283,6 +301,7 @@
 
 .floatingCardTitle {
     font-family: var(--heading-font);
+    font: var(--heading2);
```
alisey, line 304: Remove `font-family` on the previous line.

src/styles/reset.scss
alisey: It doesn't make sense here.

lib/components/color-select/color-select.module.scss
idoros: Reverted all the cursor changes

src/components/product-grid/product-grid.module.scss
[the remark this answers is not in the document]
alisey: I think it's wrong to add the class `.button` to things like icon buttons that don't have the appearance of a regular button. Because then any changes to the styling of the regular button could easily leak into those  buttons.

Also overrides like `.clearFiltersButton:not(:disabled):hover` would not be necessary if this button simply didn't include the class `.button`

src/styles/common.scss
```diff
@@ -90,6 +90,24 @@
...
 
+.transparentButton {
+    border: none;
+    background: transparent;
+    color: inherit;
+}
+.button:not(:disabled):hover {
+    background: transparent;
+    color: inherit;
+}
```
alisey, line 101: There's already a different rule for `.button:not(:disabled):hover` above in this file. And it has different styling.

idoros: Good catch. It was meant to be `.transparentButton:not(:disabled):hover`

src/components/product-grid/product-grid.module.scss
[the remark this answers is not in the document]
idoros: I can add a specific class with a something like `iconButton`?

src/styles/reset.scss
tomrav: I think Josh makes a good case for it to be in the reset in his explanation: https://www.joshwcomeau.com/css/custom-css-reset/#nine-root-stacking-context-10

Quote from site: 
Root stacking context:
This last one is optional. It's generally only needed if you use a JS framework like React.
As we saw in [“What The Heck, z-index??”(opens in new tab)](https://www.joshwcomeau.com/css/stacking-contexts/), the isolation property allows us to create a new stacking context without needing to set a z-index.
This is beneficial since it allows us to guarantee that certain high-priority elements (modals, dropdowns, tooltips) will always show up above the other elements in our application. No weird stacking context bugs, no z-index arms race.
You should tweak the selector to match your framework. We want to select the top-level element that your application is rendered within. For example, create-react-app uses a `<div id="root">`, so the correct selector is `#root`.

src/components/product-grid/product-grid.module.scss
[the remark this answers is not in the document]
alisey: I think `.iconButton` and `.linkButton` would be nice, we could provide hover and disabled styles for those.

Then for example this `clearFilters` button could be styled like:

```jsx
<button className={classNames(styles.clearFiltersButton, 'linkButton')}>
```

```diff
.clearFiltersButton {
    margin-top: 32px;
-   padding: 0;
-   cursor: pointer;
}

-.clearFiltersButton:not(:disabled):hover {
-    color: rgba(23, 17, 13, 0.7);
-}
```

src/styles/reset.scss
alisey: He's making a case for adding `isolation: isolate` to the root element, but he's not making a case for putting it in `reset.css`

We already have styles defined for this element, and I just wanted to group `isolation: isolate` with those styles here:

https://github.com/codux-templates/e-commerce-reclaim/blob/0c1446127802a417b09a76ad0b7df9225c9393c1/src/components/site-wrapper/site-wrapper.module.scss#L1-L5

Relying on application-specific element IDs in the reset.css seems wrong, but fine, keep it here.

alisey:
Default browser focus right clashing with our custom focus right:

image

https://github.com/codux-templates/e-commerce-reclaim/blob/0c1446127802a417b09a76ad0b7df9225c9393c1/src/styles/index.scss#L13-L18

I think it's better to remove the default outline in this rule instead of adding `outline: none` to individual elements, because it's easy to overlook accidentally.

src/styles/reset.scss
tomrav: I admit you've convinced me, we'll change it.

src/components/product-grid/product-grid.module.scss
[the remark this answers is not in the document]
idoros: replaced

idoros:
> I think it's better to remove the default outline in this rule instead of adding outline: none to individual elements, because it's easy to overlook accidentally:

Do you want to add `outline: none` to everything? If so, I think it would make since to move write it next to the `:focus-visible { box-shadow...}` and move them to the reset. 

However, I wonder if using `box-shadow` for focus indicators might introduce issues, since it could conflict with existing `box-shadow` styles on other elements. What do you think?

alisey:
One nice thing about `box-shadow` is that you can you can make it work well by default both on light and dark background: https://jsfiddle.net/gz0buxhj/

But outline has its own advantages, so generally I don't mind using either. I've never experienced issues with `box-shadow` for focus conflicting with regular shadows.

> Do you want to add outline: none to everything?

Yes, if we're adding box-shadow to indicate focus on all elements, we should remove the outline so we don't display both. Since we're using a site-specific color, it’s not really a "reset" but rather a "set."

idoros:
> I've never experienced issues with box-shadow for focus conflicting with regular shadows.

Simply adding a default `box-shadow` to the button style could cause it to be replaced on focus - and worse, it might completely override the intended focus indication.

Anyway, this could be improved at another time and PR, I will add a reset to the `outline` next to the focus rule.

PR #205, opened 2024-11-05
fix: make avatar image cross-origin

olehrakwix:
without `crossOrigin="anonymous"` attribute, there are issues logged in dev tools saying that request is blocked according to CORB (not CORS) policy.

looks like when loading image, browser tries to send cookies (because user is logged in in gmail etc.) and this is not allowed for cross-origin requests


alisey (commented):
Learn something new every day.

PR #206, opened 2024-11-05
feat: limit max width of cards in the featured products section

alisey:
When the featured products section contains a single image, the image takes full page width, and because of 3/4 aspect ratio it's like two screens tall and huuuuge.

See here: https://github.com/wixplosives/codux/issues/23287

I've restricted the max width to 640px. Ought to be enough for anybody.

### DIFF
--- a/app/routes/_index/route.tsx
+++ b/app/routes/_index/route.tsx
@@ -65,6 +65,7 @@ export default function HomePage() {
                 categorySlug="new-in"
                 title="New In"
                 description="Embrace a sustainable lifestyle with our newest drop-ins."
+                productCount={4}
             />
 
             <BackgroundParallax
@@ -93,6 +94,7 @@ export default function HomePage() {
                 categorySlug="best-sellers"
                 title="Best Sellers"
                 description="When quality is eco-friendly. Explore our top picks."
+                productCount={4}
             />
         </div>
     );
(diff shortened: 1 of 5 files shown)


src/components/product-card/product-card.module.scss
```diff
@@ -1,3 +1,8 @@
+.productCard {
+    display: flex;
+    flex-direction: column;
```
alisey, line 3: Note: this doesn't serve any practical purpose, this is just to have a class on the root element if users want to style it.

src/components/featured-products-section/featured-products-section.module.scss
```diff
@@ -61,7 +61,7 @@
         font-size: 14px;
     }
 
-    .productsRow {
+    .products {
```
alisey, line 64: Note: removed `row` since it can wrap to multiple lines.

app/routes/_index/route.tsx
```diff
@@ -65,6 +65,7 @@ export default function HomePage() {
                 categorySlug="new-in"
                 title="New In"
                 description="Embrace a sustainable lifestyle with our newest drop-ins."
+                productCount={4}
```
alisey, line 68: Note: explicit counts to make the Props Panel in Codux nicer.

tetianaman (approved):

PR #210, opened 2024-11-06
fix: products sorting select styles in prod

Letushev:


src/components/product-sorting-select/product-sorting-select.module.scss
```diff
@@ -1,31 +1,34 @@
-.select {
-    width: 240px;
-    justify-content: end;
-    background: none;
-    border: none;
-    font: var(--paragraph3);
-}
+/* Increasing specificity to ensure that default select styles are overridden */
```
yurii-ve, line 1: does it make sense put the link from the description here?

yurii-ve (approved):

alisey:
I'm still thinking how to do this differently.

Alternatives so far:
1. Import `select.module.scss` in the root route.
2. Explicitly "compose" with the base class name `.select.productSortingSelect {}`, like in Stylable.
3. Define two variants each having full styling: `"regular"` and `"menu"`, so we won't have to add both to the same select. This doesn't fully solve the issue: e.g. if the variant sets the `width: 100%` by default, then a specific route will have problem overriding it.

PR #211, opened 2024-11-06
feat: Tweaks to the Accordion and OrderSummary component

yurii-ve:
These changes needed for the `MyOrders` page that will come in the next PR

Accordion:
 - added `initialOpenItemIndex` prop (currently first item is opened by default)

OrderSummary:
 - UI tweaks (to be able to use the same component in the "Thank You" and "My Orders" pages)
 - added "Price" label in the order item details

Updated `OrderSummary` on the Thank You page looks like this:

image

### DIFF
--- a/app/routes/thank-you/route.module.scss
+++ b/app/routes/thank-you/route.module.scss
@@ -12,8 +12,12 @@
 }
 
 .orderNumber {
-    margin-bottom: 32px;
+    margin-bottom: 52px;
     font: var(--paragraph3);
+}
+
+.orderSummary {
+    max-width: 800px;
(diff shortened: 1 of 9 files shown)


app/routes/thank-you/route.module.scss
```diff
@@ -12,10 +12,14 @@
 }
 
 .orderNumber {
-    margin-bottom: 32px;
+    margin-bottom: 52px;
```
yurii-ve, line 15: because `OrderSummary` padding-top was removed

alisey (approved):

PR #212, opened 2024-11-07
feat: Add my orders page

yurii-ve:
UI:

Screenshot 2024-11-07 at 12 03 23

Screenshot 2024-11-07 at 12 02 43

### DIFF
--- a/app/routes/members-area.my-orders/route.module.scss
+++ b/app/routes/members-area.my-orders/route.module.scss
@@ -1,4 +1,66 @@
-.underConstructionMessage {
-    text-align: center;
+@import '@styles/variables.scss';
+
+.title {
     font: var(--heading5);
+    font-size: 26px;
+}
+
+.message {
+    margin-top: 13px;
+}
+
+.orders {
+    margin-top: 35px;
+}
+
+.orderListHeader,
+.orderHeader {
+    height: 78px;
+    display: flex;
+    align-items: center;
+    justify-content: stretch;
+    gap: 20px;
+}
+
+.orderListHeader > div,
+.orderHeader > div {
+    padding: 0 12px;
+    flex-basis: 100%;
+}
+
+.orderHeaderSectionName {
+    display: none;
+}
+
+.orderListHeader {
+    height: 35px;
+    padding-bottom: 10px;
+    padding-right: 20px;
+}
+
+.orderList {
+    border-top: 1px solid var(--primary5);
+}
+
+@media (max-width: $mobile-width) {
+    .orderListHeader {
+        display: none;
+    }
+
+    .orderHeader {
+        height: auto;
+        padding: 8px 0;
+        flex-direction: column;
+        align-items: flex-start;
+        gap: 4px;
+        font: var(--paragraph3);
+    }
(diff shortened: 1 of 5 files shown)


app/routes/members-area.my-orders/route.module.scss
```diff
@@ -1,4 +1,66 @@
...
+.orderListHeader,
+.orderHeader {
+    height: 78px;
+    display: flex;
+    align-items: center;
+    justify-content: stretch;
+    gap: 20px;
+}
+
+.orderListHeader > div,
```
olehrakwix, line 25: these styles looks strange, can we add some class name? what is that? maybe columnTitle?

app/routes/members-area.my-orders/route.tsx
```diff
@@ -1,10 +1,72 @@
...
+    const ordersResponse = await api.getOrders();
+
+    return { orders: ordersResponse.orders };
+};
+
 export default function MyOrdersPage() {
+    const { orders } = useLoaderData<typeof loader>();
+
+    const accordionItems = orders.map((order) => {
+        const formatDate = (date: Date) =>
```
olehrakwix, line 21: this might be moved out of `map` scope

app/routes/members-area.my-orders/route.tsx
```diff
@@ -1,10 +1,72 @@
...
+                        <span className={styles.orderHeaderSectionName}>Order: </span>
+                        {order.number}
+                    </div>
+                    <div>
+                        <span className={styles.orderHeaderSectionName}>Status: </span>
+                        {order.status}
+                    </div>
+                    <div>
+                        <span className={styles.orderHeaderSectionName}>Total: </span>
+                        {order.priceSummary!.total!.formattedAmount}
```
olehrakwix, line 45: maybe better to use `?` in order to not crash if there will be some case that data is undefined?

lib/ecom/types.ts
```diff
@@ -67,6 +67,10 @@ export type EcomApi = {
     getAllCategories: () => Promise<Collection[]>;
     getCategoryBySlug: (slug: string) => Promise<CollectionDetails | undefined>;
     getOrder: (id: string) => Promise<OrderDetails | undefined>;
+    getOrders: () => Promise<{
+        orders: OrderDetails[];
+        metadata?: orders.CursorPagingMetadata;
```
olehrakwix, line 72: why do we need this?

app/routes/members-area.my-orders/route.module.scss
yurii-ve: I didn't want to pollute jsx too much, as there 4 divs inside of a row.. but ok :)

lib/ecom/types.ts
yurii-ve: it's something return by the SDK and contains information needed for the pagination (`count`, `hasNext`). I'll change it to return the same info as `queryProducts` does - `{items: [], totalCount: 0}`

olehrakwix (approved):

olehrakwix (approved):

PR #213, opened 2024-11-07
fix: add loading and error states to the Cart

tetianaman:
In this PR loading and error states are added to the Cart similarly to how it is done on the Cart page PR #199 

It's hard to observe loading state because `Cart` is closed by default when user loads a webpage.
Screenshot 2024-11-07 at 14 30 56

Error state:
Screenshot 2024-11-07 at 14 27 31


src/components/cart/cart-view/cart-view.tsx
```diff
@@ -32,8 +37,19 @@ export const CartView = ({
...
+    if (isLoading) {
+        return (
+            <CartFallback>
+                <Spinner size={50} />
+            </CartFallback>
+        );
+    }
 
+    if (!cart) {
+        return error ? <CartFallback>{error}</CartFallback> : null;
```
Letushev, line 49:
```suggestion
        return <CartFallback>{error}</CartFallback>;
```

Letushev (approved):

PR #158, opened 2024-10-18
feat: implement toast notifications

tetianaman:
1. 'react-hot-toast' is used to manage toast notifications.

- 'react-hot-toast' provide us with `ToastBar` component with default styles and animations which cannot be changed using Style Panel. I've created custom `Toast` component with simple animation. 
- however 'react-hot-toast' provide `style` prop to style toasts, I used classes to keep toast responsive.  

2. The `Toast` board has been created. It contains different types of toasts to give users the ability to redesign them. They are not Variants, therefore the board is not added to the UI Kit group.
3.  The `ToastContainer` board had been added to give user possibility to play with `ToastContainer` props and `toast` function and its options.

## Demos on boards

https://github.com/user-attachments/assets/b903f27f-061c-4f46-9d54-39991fe83aff

https://github.com/user-attachments/assets/ce6bd73f-d93a-4117-9e23-539e178cbb59

PR #220, opened 2024-11-08
feat: toast notifications

alisey:
Based on #158 with many tweaks.

Usage example:
```tsx
import { toast } from '~/src/components/toast/toast';

<button onClick={() => toast.success('Yay')} />;
```

#### Toast Variants

image

#### Default position

https://github.com/user-attachments/assets/87e4e0fb-d99a-4c61-afde-fff2330da186

#### Custom position

https://github.com/user-attachments/assets/9eaac944-cf25-4c95-bacd-3c0a71fb8b19

### DIFF
--- a/src/components/toast/toast.tsx
+++ b/src/components/toast/toast.tsx
@@ -17,15 +17,19 @@ export const Toast = ({ toast }: { toast: ToastData }) => {
         custom: '',
     }[toast.type];
 
-    const className = classNames(
-        styles.root,
-        variantClassName,
-        toast.visible ? styles.enterAnimation : styles.exitAnimation,
-        toast.className,
-    );
+    const animationClassName = toast.visible ? styles.enterAnimation : styles.exitAnimation;
 
     return (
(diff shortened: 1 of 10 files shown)


src/components/toast/toast.tsx
```diff
@@ -0,0 +1,37 @@
...
+export const Toast = ({ toast }: { toast: ToastData }) => {
+    const variantClassName = {
+        success: styles.success,
+        error: styles.error,
+        blank: styles.blank,
+        loading: styles.loading,
+        custom: '',
+    }[toast.type];
+
+    const className = classNames(
```
tetianaman, line 20: If we move classNames into variable, we loose className in ElementTree.
Subjectively, I use to see a className there, so having just root `div` scares me a bit (because I know there should be a class).

Screenshot 2024-11-08 at 15 02 34

alisey: The code looks worse imo, but OK.

PR #222, opened 2024-11-08
feat: add board template with component wrapper

tetianaman:
https://github.com/user-attachments/assets/3c7faa40-cca7-4ff7-8bae-d4cf2bf9f8ec

### Note

If you use NewComponent dialog and want to see DefaultBoard template there you need to use `<ContentSlot/>`. [Read more](https://help.codux.com/kb/en/article/creating-board-templates#using-a-board-template-for-new-components)


alisey:
I've decided to add an empty div inside. I'm not sure why the 'Empty Board' template has a div, but if there's a reason for it, I guess it applies here as well.

PR #226, opened 2024-11-09
refactor: move WIX_CLIENT_ID from global scope to module scope

alisey:
In `env.d.ts`, we had a reference to a non-existent `ClientENV` type:

```ts
declare interface Window {
    ENV: ClientENV;
}
```

I've removed `window.ENV`, made `WIX_CLIENT_ID` a local variable in `ecom/api.ts`, and exported `getWixClientId()` and `setWixClientId()` to access it.


alisey:
@olehrakwix @yurii-ve I'm going to YOLO this, but please review after merge because this has a potential to break things badly.

PR #233, opened 2024-11-11
feat: reimplement the price filter using RadixUI Slider

alisey:
1. Replaced our custom RangeSlider with the RadixUI Slider.
    - It functions as either a regular slider or a range slider.

    - It has good support for keyboard input.

    - When used as a range slider, it allows thumbs to move past each other, so they don't get stuck when they overlap.

    - ⚠️ It adds a focus ring to thumbs even when using keyboard. (This is due to Chrome treating [programmatic focus](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus#focusvisible) as `:focus-visible`). IMO, it looks kinda nice.

    - ⚠️ It doesn't work correctly when rendered on a board, the wrong thumb is being dragged. This seems related to Codux [intercepting focus](https://github.com/wixplosives/codux/blob/826b06fa4ac46163d2cfdc66b123be14195355b1/packages/commons/src/preview/prevent-stealing-focus.ts#L11). Once you focus on one of the thumbs using the keyboard, everything works as expected.

2. Extracted `PriceFilter` into a component and moved the price labels from the slider component into `PriceFilter`.

3. Added debounce to the price filter when dragging the slider via keyboard.

4. Changed the price filter to update search params only on mouseup, instead of mousemove.

### Slider Demo

https://github.com/user-attachments/assets/f649cd3a-6fc4-48e0-86ca-9f868cf92b6e


Letushev (approved):
Unbelievable, it works with any number of thumbs.
Goodbye my range slider 😢

alisey:
> Unbelievable, it works with any number of thumbs.

@Letushev "Your scientists were so preoccupied with whether or not they could, they didn't stop to think if they should."

PR #241, opened 2024-11-13
refactor: json -> data (for future compat)

AviVahl:
drops the last usage of deprecated `json` and replaces it with a call to `data` (needed for the `headers` option)

we need to ensure Codux supports the `data` hook before merging this.

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -3,6 +3,12 @@ import { defineConfig } from 'vite';
 import tsconfigPaths from 'vite-tsconfig-paths';
 import { netlifyPlugin } from '@netlify/remix-adapter/plugin';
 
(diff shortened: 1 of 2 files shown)


vite.config.ts
```diff
@@ -3,6 +3,12 @@ import { defineConfig } from 'vite';
 import tsconfigPaths from 'vite-tsconfig-paths';
 import { netlifyPlugin } from '@netlify/remix-adapter/plugin';
 
+declare module '@remix-run/node' {
+    interface Future {
+        v3_singleFetch: true;
```
AviVahl, line 8: this augments the type and ensures `useLoaderData` can infer the correct type.
it is the default in current `create-remix` generated apps

PR #242, opened 2024-11-13
feat: renew wix auth token

olehrakwix:
**Summary:**
This PR renews the Wix authentication token and saves it in the session if it has expired. If the renewal process fails, it logs the user out by generating a new visitor auth token.

**Frontend Considerations:**
The frontend side token refresh is not addressed in this PR due to the following reasons:
- In order to token expire, a user would need to leave the page inactive for 4 hours (the access token’s lifetime), then resume interaction without a full page reload or opening a new tab.
- Addressing this would significantly increase template code complexity.
- Additionally, the Wix SDK automatically updates expired tokens before each request. While the updated token is not preserved, the SDK will continue to fetch a new token before every request if the token expires. This could be viewed as a potential SDK issue rather than a frontend flaw.

### DIFF
--- a/src/wix/ecom/session.ts
+++ b/src/wix/ecom/session.ts
@@ -66,6 +66,9 @@ export async function initializeEcomApiForRequest(request: Request) {
         : initializeEcomApiAnonymous();
 }
 
+// renew token in one hour before expiration. the token lifetime is 4 hours
+const TOKEN_REFRESH_MARGIN_SECONDS = 60 * 60;
+
 async function getValidAuthTokens(sessionTokens: Tokens | undefined) {
     const client = createWixClient(sessionTokens);
 
@@ -77,7 +80,10 @@ async function getValidAuthTokens(sessionTokens: Tokens | undefined) {
         effectiveTokens = sessionTokens;
 
         const currentTimeStampSeconds = Date.now() / 1000;
-        if (currentTimeStampSeconds > effectiveTokens.accessToken.expiresAt) {
+        if (
+            currentTimeStampSeconds >
+            effectiveTokens.accessToken.expiresAt - TOKEN_REFRESH_MARGIN_SECONDS
+        ) {
             try {
                 effectiveTokens = await client.auth.renewToken(sessionTokens.refreshToken);
             } catch {
--- a/src/wix/ecom/session.ts
+++ b/src/wix/ecom/session.ts
@@ -69,28 +69,24 @@ export async function initializeEcomApiForRequest(request: Request) {
 // renew token in one hour before expiration. the token lifetime is 4 hours
 const TOKEN_REFRESH_MARGIN_SECONDS = 60 * 60;
 
-async function getValidAuthTokens(sessionTokens: Tokens | undefined) {
+async function getValidAuthTokens(sessionTokens: Tokens | undefined): Promise<Tokens> {
     const client = createWixClient(sessionTokens);
 
-    let effectiveTokens: Tokens | undefined;
-
     if (sessionTokens === undefined) {
-        effectiveTokens = await client.auth.generateVisitorTokens();
-    } else {
-        effectiveTokens = sessionTokens;
+        return await client.auth.generateVisitorTokens();
+    }
 
-        const currentTimeStampSeconds = Date.now() / 1000;
-        if (
-            currentTimeStampSeconds >
-            effectiveTokens.accessToken.expiresAt - TOKEN_REFRESH_MARGIN_SECONDS
-        ) {
-            try {
-                effectiveTokens = await client.auth.renewToken(sessionTokens.refreshToken);
-            } catch {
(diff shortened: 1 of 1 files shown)


src/wix/ecom/session.ts
```diff
@@ -35,31 +35,57 @@ export { commitSession, destroySession, getSession };
...
+            try {
+                effectiveTokens = await client.auth.renewToken(sessionTokens.refreshToken);
+            } catch {
+                // generate new visitor tokens if renewing process failed
+                effectiveTokens = await client.auth.generateVisitorTokens();
+            }
+        }
+    }
+
+    return effectiveTokens;
```
VladyslavG, line 90: why don't I see a setTimout nowhere which would auto update the token when almost expired ?

olehrakwix: what are you talking about?
we need to get new value, put in session and send a new session as a set-cookie header to a client

src/wix/ecom/session.ts
```diff
@@ -35,31 +35,57 @@ export { commitSession, destroySession, getSession };
...
-    const { session } = await initializeEcomSession(request);
-    const tokens = session.get('wixSessionTokens');
-    return tokens ? initializeEcomApiWithTokens(tokens) : initializeEcomApiAnonymous();
+    const { wixSessionTokens } = await initializeEcomSession(request);
+    return wixSessionTokens
+        ? initializeEcomApiWithTokens(wixSessionTokens)
+        : initializeEcomApiAnonymous();
+}
+
+async function getEffectiveWixAuthTokens(sessionTokens: Tokens | undefined) {
```
yurii-ve, line 69: `getValidAuthTokens`?

src/wix/ecom/session.ts
```diff
@@ -35,31 +35,57 @@ export { commitSession, destroySession, getSession };
...
     // Retrieve session tokens only if the OAuth client ID matches the one associated with the session;
     // otherwise, reset the session tokens due to client ID mismatch.
-    let wixSessionTokens =
+    const wixSessionTokens =
         sessionWixClientId === wixClientId ? session.get('wixSessionTokens') : undefined;
-    let shouldUpdateSessionCookie = false;
 
-    const client = createWixClient(wixSessionTokens);
-    if (wixSessionTokens === undefined) {
+    const effectiveTokens = await getEffectiveWixAuthTokens(wixSessionTokens);
```
yurii-ve, line 46: what happens if token gets expired during the session?

olehrakwix: see comment here: https://github.com/codux-templates/e-commerce-reclaim/pull/242#discussion_r1842035310

yurii-ve (approved):

src/wix/ecom/session.ts
```diff
@@ -35,31 +35,63 @@ export { commitSession, destroySession, getSession };
...
+
+        const currentTimeStampSeconds = Date.now() / 1000;
+        if (
+            currentTimeStampSeconds >
+            effectiveTokens.accessToken.expiresAt - TOKEN_REFRESH_MARGIN_SECONDS
+        ) {
+            try {
+                effectiveTokens = await client.auth.renewToken(sessionTokens.refreshToken);
+            } catch {
+                // generate new visitor tokens if renewing process failed
```
yurii-ve, line 90: redundant comment

src/wix/ecom/session.ts
```diff
@@ -35,31 +35,62 @@ export { commitSession, destroySession, getSession };
...
+// renew token in one hour before expiration. the token lifetime is 4 hours
+const TOKEN_REFRESH_MARGIN_SECONDS = 60 * 60;
+
+async function getValidAuthTokens(sessionTokens: Tokens | undefined) {
+    const client = createWixClient(sessionTokens);
+
+    let effectiveTokens: Tokens | undefined;
+
+    if (sessionTokens === undefined) {
+        effectiveTokens = await client.auth.generateVisitorTokens();
```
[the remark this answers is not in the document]
VladyslavG: I would suggest to return early here to not put a big block of code (in else) into indentation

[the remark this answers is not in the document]
VladyslavG: also you would not then need this mutable variable at all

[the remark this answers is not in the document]
VladyslavG:
```
async function getValidAuthTokens(sessionTokens: Tokens | undefined) {
    const client = createWixClient(sessionTokens);

    if (sessionTokens === undefined) {
        return await client.auth.generateVisitorTokens();
    }

    const currentTimeStampSeconds = Date.now() / 1000;
    if (currentTimeStampSeconds > sessionTokens.accessToken.expiresAt - TOKEN_REFRESH_MARGIN_SECONDS) {
        try {
            return await client.auth.renewToken(sessionTokens.refreshToken);
        } catch {
            return await client.auth.generateVisitorTokens();
        }
    }

    return sessionTokens;
}

```

PR #246, opened 2024-11-16
feat: product grid rendering improvements

alisey:
1. Load scaled-down product images instead of the 3000×4000 originals.
2. Limit the maximum width of product cards to avoid a single card [stretching to the entire screen](https://github.com/user-attachments/assets/5fbd264a-ee61-45f8-99f4-ddc7cd2abf83).
3. Remove the fade-in effect on each card because when changing filters or sorting it adds visual noise. Old results are mixed with the new, and only new results begin animating.


src/components/featured-products-section/featured-products-section.tsx
```diff
@@ -34,7 +33,7 @@ export const FeaturedProductsSection = (props: FeaturedProductsSectionProps) =>
                           <ProductLink key={product._id} productSlug={product.slug!}>
                               <ProductCard
                                   name={product.name!}
-                                  imageUrl={product.media?.mainMedia?.image?.url}
+                                  imageUrl={getProductImageUrl(product, 700, 700)}
```
idoros, line 36: Consider using named args here to make values meaningful

src/components/product-grid/product-grid.module.scss
```diff
@@ -19,6 +19,6 @@
 
 @media (max-width: 1024px) {
     .productGrid {
-        --minCardSize: 250px;
+        grid-template-columns: repeat(auto-fit, minmax(250px, 1fr));
```
idoros, line 22: I prefer to have a single calculation and change a variable, but maybe this works better for Codux UX

src/components/featured-products-section/featured-products-section.tsx
alisey: Changed to:
```ts
getProductImageUrl(product, { minHeight: 700 })
```

src/components/product-grid/product-grid.module.scss
alisey: Changed to a pair of variables for min and max:
```css
/* The max width must be at least 3/2 of the min width to prevent gaps when
 * transitioning from 3 to 2 items per row. */
--minColumnWidth: 350px;
--maxColumnWidth: 540px;
```

alisey:
> Is the scroll still slow after restricting the images size and without the effect removal?

No, the image size was the issue. I still removed the effect, as it looked a bit too busy when changing filters or sorting. You get a mix of old and new product cards, and only some of them start animating.

PR #248, opened 2024-11-17
Barak/bi

barak007:
This pull request introduces a new analytics context to the application using `WixBiAnalytics` and integrates it into the main application component. The changes include creating a new context provider for Wix BI, updating the root component to use this provider, and adding a new dependency for the analytics library.

Open questions
1. node - because we don't have persistent `visitorId` in the server yet I currently disabled events in node.
2. dev - currently we null transport on `localhost` but we might want to be able to enable them?
3. what should happen if a site does not have metasite id


app/bi.tsx
```diff
@@ -0,0 +1,51 @@
+import React, { useEffect, useMemo } from 'react';
```
barak007, line 1: should this be a part of the template? or we want to extract it?

PR #250, opened 2024-11-18
Fix: make shipping info dynamic

tetianaman:
On the template it is shown the name of shipping region.

Screenshot 2024-11-18 at 10 05 17

User can manage shipping regions in Wix Studio backoffice:

Screenshot 2024-11-18 at 10 08 31

### DIFF
--- a/app/routes/cart/route.tsx
+++ b/app/routes/cart/route.tsx
@@ -90,7 +90,7 @@ export default function CartPage() {
                         </span>
                     </div>
                     <div className={styles.summaryRow}>
-                        <span className={styles.location}>
+                        <span className={styles.shippingInfo}>
                             {cartTotals?.shippingInfo?.region?.name}
                         </span>
                     </div>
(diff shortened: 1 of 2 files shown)


app/routes/cart/route.tsx
```diff
@@ -90,7 +90,9 @@ export default function CartPage() {
                         </span>
                     </div>
                     <div className={styles.summaryRow}>
-                        <span className={styles.location}>Poland</span>
+                        <span className={styles.shippingInfo}>
+                            {cartTotals?.shippingInfo?.region?.name}
```
Letushev, line 94: I'm not sure if it could be undefined, but don't display `summaryRow` when it is absent.

Letushev (approved):

app/routes/cart/route.tsx
tetianaman: I looks ok without data, but I've fixed it to get rid of redundant padding.

PR #253, opened 2024-11-20
feat: Use remix actions in my account page

yurii-ve:
Use Remix actions for updating user details and sending reset password email.

- added two action: `/members-area/my-account/update-details` and `/members-area/my-account/reset-password`
- updated my-account page to use forms
- removed `useAccountDetailsForm` hook


app/routes/members-area.my-account.reset-password/route.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+    const userEmail = formData.get('email') as string | undefined;
+
+    if (!userEmail) {
+        throw new Error('Missing user email to send reset password email');
+    }
+
+    await api.sendPasswordResetEmail(userEmail, request.url);
+
+    return;
```
olehrakwix, line 16: this return is not needed

app/routes/members-area.my-account.update-details/route.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+    await api.updateUser(userId, {
+        contact: {
+            firstName,
+            lastName,
+            phones: phoneNumber ? [phoneNumber] : [],
+        },
+    });
+
+    return redirect('/members-area/my-account');
```
olehrakwix, line 25: do we need this redirect? is it really reloads page? I think it is better to just do nothing

src/wix/users/use-account-details-form.ts
olehrakwix, line 1: this is sad😢😢

yurii-ve: this is life 🤷‍♂️

app/routes/members-area.my-account.update-details/route.tsx
olehrakwix: it might be an issue. when I was working on that, I found that sometimes phone number is not updated immediately. could you please double check this?

yurii-ve: otherwise we'll have `/members-area/my-account/update-details` in the url bar

olehrakwix: and also, why do we need to fetch new data? I mean, new data will be the same as we posted, so the same that is current input values, doesn't it?

app/routes/members-area.my-account.update-details/route.tsx
```diff
@@ -0,0 +1,26 @@
...
+export async function action({ request }: ActionFunctionArgs) {
+    const api = await initializeEcomApiForRequest(request);
+    const formData = await request.formData();
+
+    const userId = formData.get('userId') as string | undefined;
+    const firstName = formData.get('firstName') as string | undefined;
+    const lastName = formData.get('lastName') as string | undefined;
+    const phoneNumber = formData.get('phoneNumber') as string | undefined;
+
+    if (!userId) {
```
VladyslavG, line 13: can't you return without reading other fields if user id is missing ?

app/routes/members-area.my-account.reset-password/route.tsx
VladyslavG, line 16: why do you need return void here ?

maybe some redirect ?

app/routes/members-area.my-account/route.tsx
```diff
@@ -23,36 +22,34 @@ export async function loader({ request }: LoaderFunctionArgs) {
...
 
+    const navigation = useNavigation();
+
+    const isUpdatingUserDetails =
+        navigation.state === 'submitting' &&
+        navigation.formAction === '/members-area/my-account/update-details';
+
+    const isResettingPassword =
+        navigation.state === 'submitting' &&
+        navigation.formAction === '/members-area/my-account/reset-password';
```
VladyslavG, line 51: I don't like this full route path here. seems too brittle.
far from malleable programming experience. In the small.

app/routes/members-area.my-account.update-details/route.tsx
olehrakwix: so, it works, because after redirect loader data is updated, but it is used on our page as initial value for `setState`, so chaning that has no effect on a `useState`

olehrakwix (approved):

olehrakwix (approved):

olehrakwix (approved):

VladyslavG (approved):

PR #255, opened 2024-11-21
feat: Mock data and actions for the define app mode (by mock remix loaders and actions)

yurii-ve:
### OPTION 2
This is possible solution how to prevent login-required pages from breaking in Codux by mocking some of functionality. 

**Changes:** Added mock loaders and actions that are going to be called instead of the real ones in the define app mode.

**Downside:**  there is no feedback to the user that mocked data/action is used/performed.

**UX:** https://drive.google.com/file/d/14AKhfjcjwIG-0gEwYipgqKE4oEVvF4cI/view?usp=sharing


app/routes/members-area.my-account/codux-loader-data.ts
```diff
@@ -0,0 +1,44 @@
...
+                    _id: 'address_id',
+                },
+            ],
+            customFields: {},
+        },
+        profile: {
+            nickname: 'John Doe',
+            slug: 'jogn_d',
+            photo: {
+                url: '',
```
olehrakwix, line 31: maybe it make sense to use some mock image

app/routes/members-area.my-account/route.tsx
```diff
@@ -19,6 +24,12 @@ export async function loader({ request }: LoaderFunctionArgs) {
     return { user };
 }
 
+// will be called if app is run in Codux because fetching user details requires
+// user to be logged in but it's currently can't be done through Codux
+export async function coduxLoader(): ReturnType<typeof loader> {
+    return coduxLoaderData;
```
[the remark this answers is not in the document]
yurii-ve: maybe just `loaderMockData`?

app/routes/login/route.tsx
```diff
@@ -21,3 +21,14 @@ export async function loader({ request }: LoaderFunctionArgs) {
         },
     });
 }
+
+// will be called if app is run in Codux because logging in can be done through Codux yet
```
alisey, line 25: Fix the comment.

PR #258, opened 2024-11-26
feat: add CSS vars for page background and text color

alisey:
Allow users to easily customize the `<body>` background and text colors through the global variables panel, as the style panel currently does not support editing the `<body>`.

image

### DIFF
--- a/src/components/avatar/avatar.module.scss
+++ b/src/components/avatar/avatar.module.scss
@@ -1,7 +1,6 @@
 .root {
     width: 22px;
     height: 22px;
-    color: var(--primary5);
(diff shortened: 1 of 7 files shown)


src/components/avatar/avatar.module.scss
```diff
@@ -1,7 +1,6 @@
 .root {
     width: 22px;
     height: 22px;
-    color: var(--primary5);
```
alisey, line 4: Note: this was redundant since it should simply inherit the text color.

tetianaman (approved):

PR #259, opened 2024-11-28
feat: implement search functionality

AviVahl:
- searches product names from all products
- uses existing all-products route to show results and allow further sorting/filtering
- leverages `$contains` (no fuzzy yet), which is publicly available on the rest API. the JS API implements it, but does not expose it via the types.
- fuzzy is easily accessible in V3 API

### DIFF
--- a/app/routes/products.$categorySlug/route.tsx
+++ b/app/routes/products.$categorySlug/route.tsx
@@ -140,7 +140,9 @@ export default function ProductsPage() {
 
                 <div className={styles.main}>
                     <div className={styles.categoryHeader}>
-                        <h1 className={styles.categoryName}>{category.name}</h1>
+                        <h1 className={styles.categoryName}>
+                            {appliedFilters.search ? `"${appliedFilters.search}"` : category.name}
+                        </h1>
                         {category.description && (
                             <p className={styles.categoryDescription}>{category.description}</p>
                         )}
--- a/app/routes/products.$categorySlug/route.tsx
+++ b/app/routes/products.$categorySlug/route.tsx
@@ -143,7 +143,7 @@ export default function ProductsPage() {
                         <h1 className={styles.categoryName}>
                             {appliedFilters.search ? `"${appliedFilters.search}"` : category.name}
                         </h1>
-                        {category.description && (
+                        {category.description && !appliedFilters.search && (
                             <p className={styles.categoryDescription}>{category.description}</p>
                         )}
                     </div>
--- a/src/components/header/header.tsx
+++ b/src/components/header/header.tsx
@@ -1,9 +1,10 @@
-import { Link } from '@remix-run/react';
+import { Link, useNavigate } from '@remix-run/react';
 import classNames from 'classnames';
 import { useState } from 'react';
 import { CartIcon, MenuIcon } from '~/src/components/icons';
 import { getCartItemCount, useCartData, useCartOpen } from '~/src/wix/cart';
 import { NavigationMenu } from '../navigation-menu/navigation-menu';
+import { SearchInput } from '../search-input/search-input';
 import { SidebarNavigationMenu } from '../sidebar-navigation-menu/sidebar-navigation-menu';
 import { UserMenu } from '../user-menu/user-menu';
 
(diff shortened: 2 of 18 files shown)


AviVahl:
made additional changes after feedback from @Dana-Levy:
- added the clear icon
- ensure correct width in desktop/tablet/mobile
- adjusted font size
- ensure min 2 characters

app/routes/products.$categorySlug/route.tsx
```diff
@@ -140,7 +140,9 @@ export default function ProductsPage() {
 
                 <div className={styles.main}>
                     <div className={styles.categoryHeader}>
-                        <h1 className={styles.categoryName}>{category.name}</h1>
+                        <h1 className={styles.categoryName}>
+                            {appliedFilters.search ? `"${appliedFilters.search}"` : category.name}
```
alisey, line 144: I don't think the category description from "All products" should be shown if we have a search term, as it's unclear what this description relates to:

image

alisey: I think we also have a UX issue here. We show the search term in three places
1. The search box
2. The heading
3. The list of allied filters

Pressing "x" in the filter list clears the search term, but pressing "x" in the search box doesn't, which feels weird. So I think we either shouldn't persist the search term in the input, or "x" should do the same action in both cases.

@Dana-Levy

src/components/header/header.tsx
```diff
@@ -16,6 +17,16 @@ export interface HeaderProps {
 export const Header = ({ className }: HeaderProps) => {
     const cart = useCartData();
     const cartOpener = useCartOpen();
+    const navigate = useNavigate();
+
+    const onSubmit: React.FormEventHandler<HTMLFormElement> = (event) => {
```
alisey, line 22: It would be nicer if instead of the raw event we received the search term that is already validated. Keeps the Header component cleaner, and avoids repetition if the `SearchInput` is used in multiple places.

```ts
const onSearchSubmit = (search: string) => {
    navigate(`/products/all-products?search=${encodeURIComponent(search)}`);
};

<SearchInput onSubmit={onSearchSubmit} />
```

src/components/icons/clear-icon.tsx
```diff
@@ -0,0 +1,8 @@
+export const ClearIcon = (props: React.SVGProps<SVGSVGElement>) => (
```
alisey, line 1: 1. This is basically a smaller version of existing `CloseIcon`. Let's call it `CloseSmallIcon`.

Or rename both to `CrossIcon` / `CrossSmallIcon` to name them based on appearance rather than based on how they're used. Because otherwise we'll have a bunch of identical looking icons named Close, Delete, Clear, Dismiss, Error.

app/routes/products.$categorySlug/route.tsx
AviVahl: > I don't think the category description from "All products" should be shown if we have a search term, as it's unclear what this description relates to

removed description.

src/components/icons/clear-icon.tsx
AviVahl: renamed

src/components/search-input/search-input.module.scss
```diff
@@ -0,0 +1,46 @@
+.form {
+    position: relative;
+    left: 1em;
+    width: 16em;
+
+    @media (max-width: 768px) {
+        width: 10em;
+    }
```
alisey, line 8: This makes it not a general-purpose search input, but a search input for the specific header design.

So let's move this part along with `left: 1em; and width: 16em;` into the header CSS (my preferred option), or call this component HeaderSearchInput to make it more clear that you're not supposed to use it for anything else.

src/components/header/header.tsx
AviVahl: done. much better

src/components/search-input/search-input.module.scss
```diff
@@ -0,0 +1,46 @@
...
+.input {
+    width: 100%;
+    padding: 0 1.5em;
+    border: none;
+    background: none;
+    font-size: 0.9em;
+
+    &::placeholder {
+        font-size: inherit;
+    }
```
alisey, line 20: Does this have any effect?

src/components/search-input/search-input.module.scss
AviVahl: moved

src/components/search-input/search-input.module.scss
AviVahl: apparently not. missed it while was fiddling to match Dana's expectations.
deleted this rule

src/components/search-input/search-input.module.scss
```diff
@@ -0,0 +1,46 @@
...
+}
+
+.searchIcon {
+    position: absolute;
+    top: 50%;
+    left: 0;
+    transform: translateY(-50%);
+    display: inline-block;
+}
```
alisey, line 23: Nitpick, but I think this would be easier to achieve (and maybe easier for users to style) if implemented using flexbox instead of absolute positioning.

Something like this: https://jsfiddle.net/1ynr2ect/

alisey: Oh, and wrapping everything with a label so the input receives focus if you click on the search icon or the "x" icon would be a nice touch.

alisey (approved):

src/components/search-input/search-input.module.scss
AviVahl: wrapped with label. used flex.

PR #263, opened 2024-12-05
Shiree/development

ShireeWill:
- Changing Typography from overrides to use of variants.
- Small UI fixes


alisey:
Continued in #271

PR #271, opened 2024-12-18
feat: unify typography

alisey:
Based on #263, but without changes that are not related to typography.

### DIFF
--- a/src/components/header/header.module.scss
+++ b/src/components/header/header.module.scss
@@ -13,8 +13,7 @@
 }
 
 .logo {
-    font: var(--heading3);
-    font-size: 26px;
+    font: var(--heading4);
 }
 
 .searchInput {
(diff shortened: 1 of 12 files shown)


src/components/header/header.module.scss
```diff
@@ -104,6 +103,6 @@
         flex-direction: column-reverse;
         align-items: start;
         justify-content: center;
-        font-size: 13px;
+        font: var(--paragraph3);
```
alisey, line 106: This is redundant, it's the same as in desktop mode.

src/styles/utils.scss
```diff
@@ -190,7 +190,7 @@
     }
 
     .textPage h2 {
-        font-size: 20px;
+        font-size: 18px;
```
alisey, line 193: Is this an intentional use of override?

src/styles/utils.scss
```diff
@@ -307,7 +307,7 @@
 
 @container (max-width: 300px) {
     .linkCardTitle {
-        font-size: 30px;
+        font: var(--heading4);
```
alisey, line 310: Redundant, same as on desktop.

